PR #6, opened 2022-04-22
ci: use pnpm version from package.json#engines

nathanaelhoun:
And separe deps installation and packaging into two steps


nathanaelhoun:
No longer useful

Issue #8, opened 2022-05-02
[Bug]: Outlook service does not properly save credentials


SpecialAro:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Ferdium Version

6.0.0-nightly.22

### Recipe Name

Office 365 Outlook

### Recipe Version

1.6.0

### Steps to reproduce

1. Add Outlook 365 recipe
2. Login to your account
3. Close Ferdium
4. Open Ferdium
5. Login screen comes up again (shouldn't be necessary to login again)

### Expected Behavior

1. Add Outlook 365 recipe
2. Login to your account
3. Close Ferdium
4. Open Ferdium
5. Account logged in automatically

### Actual Behavior

Account is not being logged in automatically after closing and opening Ferdium.

Sauceee:
Found a workaround: With the selfhosted option and URL: https://outlook.live.com/mail/

SpecialAro:
> Found a workaround: With the selfhosted option and URL: https://outlook.live.com/mail/
Hello @Sauceee , yes, that's the workaround I've been using. We should migrate to that link in the future on the recipe's code.

dchmelik:
I have same problem with some Outlook (personal, not school) but not other MS (Teams).

PR #9, opened 2022-05-02
Fix outlook issue by changing the service URL

SpecialAro:
This solves https://github.com/ferdium/ferdium-recipes/issues/8.

We previously had two recipes for outlook (one normal, and another for 365) - does this change affect anything on that?

### DIFF
--- a/recipes/office365-owa/package.json
+++ b/recipes/office365-owa/package.json
@@ -1,10 +1,11 @@
 {
   "id": "office365-owa",
   "name": "Office 365 Outlook",
-  "version": "1.6.0",
+  "version": "1.6.1",
   "license": "MIT",
   "aliases": [
-    "live.com"
+    "live.com",
+    "outlook"
   ],
   "config": {
     "serviceURL": "https://outlook.live.com/mail/",
--- a/recipes/office365-owa/package.json
+++ b/recipes/office365-owa/package.json
@@ -7,7 +7,7 @@
     "live.com"
   ],
   "config": {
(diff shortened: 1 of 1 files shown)


vraravam (changes requested):
Please run `pnpm run package` - that will fail since the recipe version needs a bump in the corresponding `package.json`

vraravam:
Also, since we are anyways editing this service recipe, could you please add the "outlook" as an alias in `package.json`?

vraravam:
I had a chance to try this on my client's outlook (that uses okta for SSO) - but, this change did not have any impact. Once Ferdium is closed and reopened, I was taken back to the okta sign-in page.

SpecialAro:
> Please run `pnpm run package` - that will fail since the recipe version needs a bump in the corresponding `package.json`
image

This didn't fail... But it should, because I didn't upgrade the json version, yes.

> Also, since we are anyways editing this service recipe, could you please add the "outlook" as an alias in package.json?
Next commit will have it.

> I had a chance to try this on my client's outlook (that uses okta for SSO) - but, this change did not have any impact. Once Ferdium is closed and reopened, I was taken back to the okta sign-in page.
Should we close this PR, given the fact that the issue may be other?

vraravam:
>  Should we close this PR, given the fact that the issue may be other?

Yes, im not sure if this is a valid change - for the intended goal.
Also, the old url works similarly from my testing. So, I'm not sure about the need for this PR

vraravam:
Closing as per discussion with @SpecialAro

PR #48, opened 2022-06-07
[Gmail] add support for priority inbox

bernikr:
in gmail there are settings for different styles of inboxes. if this is set to "priority inbox" or "important first", the unread count consists of two numbers:

image

The first number is the number of unread "important" messages and the second number is the number of all unread messages.

Before Ferdium would read a wrong count and report this situation as 12 unread messages which is clearly wrong. I decided to change that so that the first number is reflected in the unread message count and the second number is used for the "indirect message" count

### DIFF
--- a/recipes/gmail/package.json
+++ b/recipes/gmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "gmail",
   "name": "Gmail",
-  "version": "1.5.0",
+  "version": "1.6.0",
   "license": "MIT",
   "config": {
     "serviceURL": "https://mail.google.com"
--- a/recipes/gmail/package.json
+++ b/recipes/gmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "gmail",
   "name": "Gmail",
-  "version": "1.6.0",
+  "version": "1.5.1",
   "license": "MIT",
   "config": {
     "serviceURL": "https://mail.google.com"
--- a/recipes/gmail/webview.js
+++ b/recipes/gmail/webview.js
@@ -15,8 +15,8 @@ module.exports = Ferdium => {
   }
 
   const getMessages = () => {
-    let count = 0;
-    let count2 = 0;
+    let countImportant = 0;
+    let countNonImportant = 0;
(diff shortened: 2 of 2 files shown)


recipes/gmail/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "id": "gmail",
   "name": "Gmail",
-  "version": "1.5.0",
+  "version": "1.6.0",
```
SpecialAro, line 4: This should be `1.5.1` and not `1.6.0`

recipes/gmail/webview.js
```diff
@@ -16,7 +16,7 @@ module.exports = Ferdium => {
 
   const getMessages = () => {
     let count = 0;
-
+    let count2 = 0;
```
SpecialAro, line 19: Can we name both of these accordingly? I suggest those:
`count -> countImportant`
`count2 -> countNonImportant`

recipes/gmail/webview.js
```diff
@@ -25,14 +25,23 @@ module.exports = Ferdium => {
...
+              );
+            }
           }
         }
       }
     }
-
-    // set Ferdium badge
-    Ferdium.setBadge(count);
+    Ferdium.setBadge(count, count2);
```
SpecialAro, line 44: Have you tested this locally?

SpecialAro (changes requested):
Hi @bernikr, thank you for your contribution!

Overall, it all looks good to me. Nevertheless, I haven't tested this locally. Have you done so?

If you can, please review the comments I left and solve them, so this can be merged. Thank you!

recipes/gmail/webview.js
bernikr: yes i have tested this.

If there are important notification there number gets shown in a red bubble, if not but there is at least one not important notification a blue dot get shown.

vraravam (approved):

SpecialAro (approved):

vraravam:
@all-contributors please add @bernikr for code

PR #57, opened 2022-06-15
Updated dark mode for 'messenger' service

vraravam:
Changes copied from https://github.com/ducfilan/Dark-mode-Franz-Ferdi (attribution already present in the file)

I haven't tested this since I dont use FB. Please test and then approve.


SpecialAro (approved):
LGTM 😄 Tho, I'm having a problem on disabling the darkmode (it enables itself automatically). But this maybe caused by something on my local.

Issue #50, opened 2022-06-09
[Recipe Request]: Proton Drive


haykh:
### Preflight Checklist

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Name of the service

Proton Drive

### URL of the service

drive.proton.me

### GitHub URL to a recipe that has already been created by someone else or you

https://github.com/haykh/not-a-repository/tree/master/6-8-2022/protondrive

### Features you want implemented

So apparently Proton is undergoing a huge rebranding, and they've added Proton Calendar and Drive to their services. Proton Drive only has a web-app currently, so I think it's very much worth adding them to the services (see repo). Might also be worth changing the Proton Mail logo.

vraravam:
@haykh - if you can specify the MIT license for that recipe, I can copy it over for use in Ferdium by all users. Sorry for the delay, somehow I didn't get notifications for this repository's tickets.

vraravam:
Belay that, its specified as "MIT" in the `package.json` - that works for me

vraravam:
Actually, rather than me copying each file into a recipe's PR, can you just create a PR against this repo? That way, you get attributed to the contribution!

PR #61, opened 2022-06-16
🍱 protondrive added

haykh:
Solves #50


recipes/protondrive/webview.js
```diff
@@ -0,0 +1,8 @@
+const _path = _interopRequireDefault(require('path'));
+
+function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
+
+module.exports = (Ferdium) => {
```
vraravam, line 5: If the config setting in `package.json` has `hasIndirectMessages` as `true`, then there should be some corresponding code in this module to call `Ferdium.setBadge()`. Without such a call, the badges won't appear. Is that ok?

vraravam (changes requested):

haykh:
Should be fine. I'm actually not sure if they ever send any browser notifications, never seen one. If this changes -- I'll update it accordingly.

recipes/protondrive/webview.js
haykh: Should be fine. I'm actually not sure if they ever send any browser notifications, never seen one. If this changes -- I'll update it accordingly.

vraravam (changes requested):

vraravam (approved):

vraravam:
@haykh - please run `pnpm package` to check and fix the above issue.

vraravam:
@all-contributors please add @haykh for code

PR #80, opened 2022-06-25
New recipe: 'pushover'

vraravam:


vraravam:
@bonelifer - I have copied over the MIT-licensed recipe - and changed it to conform to the updated structure expected by Ferdium. I haven't verified that the url and the unread count badges are working (I am not going to signup for this service just to test it out). I will merge this PR, and it should appear in the next nightly. Please test it in the new version. TIA

PR #82, opened 2022-06-27
Updated FB messenger url

vraravam:


Shadester:
Why was this changed?

vraravam:
@Shadester : Based on this: https://github.com/TheGoddessInari/hamsket/commit/86c1fdeae14009d39282ab841305da4a48083358

Shadester:
Oh. So to support darkmode, this had to be done. I would definitely add a separate recipe for that in that case, since I really only want to have messenger and not the facebook stuff that comes with the dark mode enabled one.

image

image

Shadester:
If someone wants to have darkmode in messenger, they can use the facbook service and expand to only show messenger.

scholtzm:
I agree, this should have been a separate recipe. Just opened new can of worms with this.

vraravam:
We have reverted the above change - it should come in the next nightly in a few hours.

PR #85, opened 2022-06-29
Reverted fb messenger url; Add new recipe for FB messages

vraravam:
I have reused the same icon and unread count logic for fb-messages from messenger.
Since I have blocked FB from my hosts file and I dont use these services, I dont know if the icon needs to be updated. Would request one of the other contributors to update the same.


cino (approved):
Looks good to me, can't test as well as not having a FB account

SpecialAro:
@vraravam LGTM, but what's the need of having FB Messenger? Messenger is exactly the same... Adding that recipe is like havving the same as Facebook recipe.

Therefore, I suggest we only revert the messenger URL and don't add FB messages

SpecialAro (changes requested):
@vraravam LGTM, but what's the need of having FB Messenger? Messenger is exactly the same... Adding that recipe is like havving the same as Facebook recipe.

Therefore, I suggest we only revert the messenger URL and don't add FB messages

vraravam:
ok - i can remove the new recipe from the PR - give me a few secs

PR #96, opened 2022-07-03
Explicitly use user agent of chrome for discord receipe. Fixes #31

### LINKED ISSUE #31: [Bug]: Discord app does not show the search box when used within ferdi
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Ferdium Version

6.0.0-nightly.39

### Recipe Name

discord

### Recipe Version

Latest (1.7.2)

_No response_

### Steps to reproduce

1. Open ferdium discord channel within ferdi.

### Expected Behavior

The search box should be visible

### Actual Behavior

The search box is not visible

### Screenshots

# When Opened in Ferdi
Screenshot 2022-05-18 at 8 58 47 AM

# When Opened in firefox
Screenshot 2022-05-18 at 8 58 03 AM

### Additional Information

_No response_

kishaningithub:
Fixes #31

### DIFF
--- a/recipes/discord/index.js
+++ b/recipes/discord/index.js
@@ -1,5 +1,5 @@
 module.exports = Ferdium => class Discord extends Ferdium {
   overrideUserAgent() {
-    return 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36';
+    return window.navigator.userAgent.replace("Apple Mac OS X", "Intel Mac OS X");
   }
 };
--- a/recipes/discord/index.js
+++ b/recipes/discord/index.js
@@ -1,5 +1,8 @@
 module.exports = Ferdium => class Discord extends Ferdium {
   overrideUserAgent() {
-    return window.navigator.userAgent.replace("Apple Mac OS X", "Intel Mac OS X");
+    return window.navigator.userAgent.replace('(KHTML, like Gecko)', '(KHTML, like Gecko) discord/0.0.250')
+      .replace('Electron', 'Discord')
+      .replace('Ferdium', 'Discord')
+      .replace('Apple Mac OS X', 'Intel Mac OS X');
   }
 };
--- a/recipes/discord/index.js
+++ b/recipes/discord/index.js
@@ -1,8 +1,5 @@
 module.exports = Ferdium => class Discord extends Ferdium {
   overrideUserAgent() {
-    const useragent = window.navigator.userAgent;
-    const parts = useragent.split('(KHTML, like Gecko)');
-
-    return parts.join('(KHTML, like Gecko) discord/0.0.250').replace('Electron', 'Discord').replace('Ferdium', 'Discord');
+    return 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36';
   }
 };
--- a/recipes/discord/package.json
+++ b/recipes/discord/package.json
@@ -1,7 +1,7 @@
 {
   "id": "discord",
   "name": "Discord",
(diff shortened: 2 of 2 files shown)


vraravam:
@kishaningithub - the user can override the `user-agent` from the service-specific settings screen - why do we need this PR?

kishaningithub:
One thing I am not sure is, what is the percentage of user base facing this problem. If it is small i feel service specific setting you mentioned should be enough, if not i feel having this in the recipe would be a sensible default.

vraravam:
> One thing I am not sure is, what is the percentage of user base facing this problem. If it is small i feel service specific setting you mentioned should be enough, if not i feel having this in the recipe would be a sensible default.

True - but, then again, avoiding to hard-code this string was the reason why this feature (preference UI) was implemented in the first place. Also, having the hard-coded string would mean lock-in for future until a developer goes and manually changes it in the code - which I am loathe to do.

kishaningithub:
@vraravam Agreed. I did not like the hard coding part myself.

However, In search of a generic solution, was digging in into what was the difference that was causing this. Thanks to the developer tools available within ferdium(Both generic and service specific ones) the experience was a breeze. Thanks a lot for that!

The user agent that was getting generated by ferdium out of the box was
```
Mozilla/5.0 (Macintosh; Apple Mac OS X 12_4_0) AppleWebKit/537.36.0 (KHTML, like Gecko) Chrome/102.0.5005.134 Safari/537.36.0
```

The user agent generated by chrome is
```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/103.0.0.0 Safari/537.36
```

Upon looking closely at the above i figured the only diff was within the OS part, Ferdium was saying it was "Apple Mac OSX" where as chrome was saying it was "Intel Mac OS X" (Wierdly this is the case even with a specific arm64 build of chrome)

Also was looking into a compilation of user agents for chrome and electron

Chrome user agents: https://www.whatismybrowser.com/guides/the-latest-user-agent/chrome
Electron user agents: https://developers.whatismybrowser.com/useragents/explore/software_name/electron-application/

Nowhere i see "Apple Mac OSX" for mac os builds. Any pointers as to where the default user agents are set within ferdium-app would be helpful :-)

For now, Have modified the PR to replace the OS part  from "Apple Mac OSX" to "Intel Mac OSX". Have tested this and it works beautifully

recipes/discord/index.js
```diff
@@ -1,8 +1,5 @@
 module.exports = Ferdium => class Discord extends Ferdium {
   overrideUserAgent() {
-    const useragent = window.navigator.userAgent;
-    const parts = useragent.split('(KHTML, like Gecko)');
-
-    return parts.join('(KHTML, like Gecko) discord/0.0.250').replace('Electron', 'Discord').replace('Ferdium', 'Discord');
+    return window.navigator.userAgent.replace("Apple Mac OS X", "Intel Mac OS X");
```
vraravam, line 3: You will probably also want the older string-replaces in place. Please add your `.replace()` calls as additive - so that other OSes, chip-architectures, etc continue to work.

vraravam (changes requested):

recipes/discord/index.js
vraravam: @kishaningithub - I have made the required changes (including bumping up the recipe version number) - please verify once.

kishaningithub: Tested it in my ARM M1 mac.. The change works flawlessly..

vraravam:
@all-contributors please add @kishaningithub for code

PR #112, opened 2022-07-08
Fix in app URLs for discord

SpecialAro:
Since we implemented `trapLinkClicks` we broke some of discord functionalities (such as processing its internal links correctly - which now open in the browser instead).

This fixes that issue.

Windows


vraravam (approved):
LGTM! Nice simple solution!

PR #120, opened 2022-07-12
Change isImage to link and stayInsideDiscord

SpecialAro:
#### Pre-flight Checklist

1. Please remember that if you are logging a bug for some service that has _stopped working_ or is _working incorrectly_, please log the bug [here](https://github.com/ferdium/ferdium-recipes/issues)
2. If you are requesting support for a **new service** in Ferdium, please log it [here](https://github.com/ferdium/ferdium-recipes/pulls)
3. Please remember to read the [self-help documentation](https://github.com/ferdium/ferdium-app#troubleshooting-recipes-self-help) - in case it helps you unblock yourself for issues related to older versions of recipes that were installed on your machine. (These will get automatically upgraded when you upgrade to the newer versions of Ferdium, but to get new recipes between Ferdium releases, this documentation is quite useful.)
4. Please ensure you've completed all of the following.

- [x] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues) for a feature request that matches the one I want to file, without success.

#### Description of Change
Change isImage to URL and stayInsideDiscord to handle 2 base URLs (discord.com and discordapp.com)

#### Motivation and Context
- Handle 2 base URLs (discord.com and discordapp.com)
- Improve isImage function
This PR needs to be merged **AFTER** or together with https://github.com/ferdium/ferdium-app/pull/470.

#### Checklist

- [x] My pull request is properly named
- [x] I tested/previewed my changes locally

### DIFF
--- a/recipes/discord/webview.js
+++ b/recipes/discord/webview.js
@@ -33,9 +33,11 @@ module.exports = (Ferdium, settings) => {
 
     if (link || button) {
       const url = link ? link.getAttribute('href') : button.getAttribute('title');
-      const stayInsideDiscord = url.includes('https://discordapp.com/channels/');
+      const stayInsideDiscord =
+        url.includes('https://discordapp.com/channels/') ||
+        url.includes('https://discord.com/channels/');
 
-      if (!Ferdium.isImage(url) && !stayInsideDiscord) {
+      if (!Ferdium.isImage(link) && !stayInsideDiscord) {
         event.preventDefault();
         event.stopPropagation();
 
--- a/recipes/discord/webview.js
+++ b/recipes/discord/webview.js
@@ -33,9 +33,12 @@ module.exports = (Ferdium, settings) => {
 
     if (link || button) {
       const url = link ? link.getAttribute('href') : button.getAttribute('title');
-      const stayInsideDiscord =
-        url.includes('https://discordapp.com/channels/') ||
-        url.includes('https://discord.com/channels/');
+      const skipDomains = [/^https:\/\/discordapp\.com\/channels\//i, /^https:\/\/discord\.com\/channels\//i];
+
+      skipDomains.every(skipDomain => {
+        stayInsideDiscord = skipDomain.test(url);
+        return !stayInsideDiscord;
+      });
 
       if (!Ferdium.isImage(link) && !stayInsideDiscord) {
         event.preventDefault();
--- a/recipes/zoom/webview.js
+++ b/recipes/zoom/webview.js
@@ -33,8 +33,8 @@ module.exports = (Ferdium, settings) => {
 
     if (link || button) {
       const url = link ? link.getAttribute('href') : button.getAttribute('title');
-      
-      if (!Ferdium.isImage(url)) {
+
+      if (!Ferdium.isImage(link)) {
         event.preventDefault();
         event.stopPropagation();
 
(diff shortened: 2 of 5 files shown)


recipes/zoom/webview.js
```diff
@@ -33,8 +33,8 @@ module.exports = (Ferdium, settings) => {
 
     if (link || button) {
       const url = link ? link.getAttribute('href') : button.getAttribute('title');
-      
-      if (!Ferdium.isImage(url)) {
+
+      if (!Ferdium.isImage(link)) {
```
vraravam, line 37: What happens if the user clicked on a button (and the `link` variable is undefined)?

recipes/discord/webview.js
```diff
@@ -33,9 +33,11 @@ module.exports = (Ferdium, settings) => {
 
     if (link || button) {
       const url = link ? link.getAttribute('href') : button.getAttribute('title');
-      const stayInsideDiscord = url.includes('https://discordapp.com/channels/');
+      const stayInsideDiscord =
+        url.includes('https://discordapp.com/channels/') ||
+        url.includes('https://discord.com/channels/');
 
-      if (!Ferdium.isImage(url) && !stayInsideDiscord) {
+      if (!Ferdium.isImage(link) && !stayInsideDiscord) {
```
vraravam, line 40: What happens if the user clicked on a button (and the `link` variable is undefined)?

vraravam (changes requested):

recipes/zoom/webview.js
SpecialAro: Good question. Maybe this should only be checked if it is NOT a button?

SpecialAro: Inside Ferdium.IsImage we make a check and return false if "link" is undefined

recipes/discord/webview.js
SpecialAro: Inside Ferdium.IsImage we make a check and return false if "link" is undefined

recipes/discord/webview.js
```diff
@@ -33,9 +33,14 @@ module.exports = (Ferdium, settings) => {
 
     if (link || button) {
       const url = link ? link.getAttribute('href') : button.getAttribute('title');
-      const stayInsideDiscord = url.includes('https://discordapp.com/channels/');
+      const skipDomains = [/^https:\/\/discordapp\.com\/channels\//i, /^https:\/\/discord\.com\/channels\//i];
 
-      if (!Ferdium.isImage(url) && !stayInsideDiscord) {
+      skipDomains.every(skipDomain => {
```
vraravam, line 38: i think you want to break out of the loop as soon as any of the discord domains are hit. The current loop will take the last value for `stayInsideDiscord`

vraravam (changes requested):

recipes/discord/webview.js
vraravam: @SpecialAro - pls see this comment

SpecialAro: I've tested it and, as is, it will break as soon as `!stayInsideDiscord` is false, so there's no issue (I was surprised as well with this ahah)

Gaboris: > I've tested it and, as is, it will break as soon as `!stayInsideDiscord` is false, so there's no issue (I was surprised as well with this ahah)

Yep, I did my research @vraravam. I picked every() specifically because the second it gets a false return it breaks the cycle so this way we could do it in fewer steps and less chances of bugs. 😁

PR #131, opened 2022-07-23
fix pnpm run create directions

langdon:
per https://github.com/getferdi/recipes/blob/master/docs/integration.md and testing, I think it should be `pnpm run create <recipe>`. However, I don't know pnpm particularly well.


vraravam (approved):
Nice catch @langdon ! LGTM!

vraravam:
@all-contributors pls add @langdon for docs

PR #138, opened 2022-08-04
fix: add warning to whatsapp theming

SpecialAro:
This PR fixes an issue with Whatsapp Native Theming when users try to change it using the Whatsapp UI (and not through Ferdium). This issue is described in https://github.com/ferdium/ferdium-app/issues/490 and https://github.com/ferdium/ferdium-app/issues/535.

Now, users will have a popup window like shown in the image whenever they try to change the theme through the whatsapp UI.

image

### DIFF
--- a/recipes/whatsapp/webview.js
+++ b/recipes/whatsapp/webview.js
@@ -4,6 +4,45 @@ function _interopRequireDefault(obj) {
   return obj && obj.__esModule ? obj : { default: obj };
 }
 
+function createElement(messageBody, idString) {
+  const messageText = document.createTextNode(messageBody);
+  const message = document.createElement("p");
+  message?.setAttribute("id", idString);
+  message?.setAttribute("class", idString);
+  message?.appendChild(messageText);
+
+  return message;
+}
+
+function addThemeMessage() {
+  const idString = 'ferdium-theme-message';
+  const idStringTitle = 'ferdium-theme-title';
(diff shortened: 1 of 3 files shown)


recipes/whatsapp/webview.js
```diff
@@ -48,6 +87,7 @@ module.exports = Ferdium => {
   const loopFunc = () => {
     getMessages();
     getActiveDialogTitle();
+    addThemeMessage();
```
vraravam, line 90: Will this keep popping up every second and be a bad UX?

recipes/whatsapp/webview.js
```diff
@@ -4,6 +4,45 @@ function _interopRequireDefault(obj) {
...
+  const elementExists = document.querySelectorAll(`.${idString}`)[0] || document.querySelectorAll(`.${idStringTitle}`)[0] ? true : false;
+
+  if (!elementExists) {
+    const themePopupDiv = document.querySelectorAll("._2Nr6U")[0];
+
+    // Create Ferdium Warning title element
+    const messageTitleString = 'FERDIUM WARNING!';
+    const messageTitleElement = createElement(messageTitleString, idStringTitle);
+
+    // Create Ferdium Waring message element
```
vraravam, line 29: typo

recipes/whatsapp/webview.js
SpecialAro: The function will trigger every second (or every time frame we have it implemented). But the function will only make changes as per this line (22) `if (!elementExists)`

vraravam (approved):
LGTM!

PR #145, opened 2022-08-07
Simplify getting badge count

beerpiss:
Facebook Messenger had a UI change which included adding an unread count:  
image

This PR takes advantage of this count instead of counting the badges manually.


vraravam (changes requested):
please update the version of the recipe in the recipe's `package.json` file.

vraravam:
Did you run the `pnpm package` command before staging the changes?

vraravam (approved):

vraravam:
I dont use FB, and so couldn't test this. The code looks ok. Can I request someone else to also test this on their system before this gets merged?

fschaupp:
Hi @beerpiss!
I tried your patch, but haven't received the UI change yet. So now the message count is gone for me.
Can you please create for the transition phase a fallback to the old way of reading the number?
Maybe something like 
```
if (elem) {
  // nice new stuff here
} else {
  // try the old way here
}
```

DominusExult:
I'm not getting any FB Messenger badges anymore. If need be I can check the appropriate nightly before this was added (I only learnt a few days ago about the Ferdi drama and switched from that to Ferdium 6.1 and now the last nightly)

vraravam:
without any change in our code, of the badges stop working, then its FB who have changed the css identifiers of the page. its not something we can control.

DominusExult:
> without any change in our code, of the badges stop working, then its FB who have changed the css identifiers of the page. its not something we can control.

To be more clear, I switched from Ferdi 5.8.1 to Ferdium 6.1.0 and Messenger stopped giving badges. Tried Ferdium 6.0 - same, no badges.
Retried a pristine Ferdi 5.8.1 and the badges are still there.

vraravam:
> > without any change in our code, of the badges stop working, then its FB who have changed the css identifiers of the page. its not something we can control.
> 
> To be more clear, I switched from Ferdi 5.8.1 to Ferdium 6.1.0 and Messenger stopped giving badges. Tried Ferdium 6.0 - same, no badges. Retried a pristine Ferdi 5.8.1 and the badges are still there.

you can't compare Ferdi to Ferdium - Ferdium has gone past Ferdi in its internals in a major manner by this point. When I mentioned "our code", i meant 2 different versions of Ferdium

DominusExult:
Right, I'll post it as an issue then. For whoever that looks at it, it might be interesting to know that code from 5 months ago still shows the badges, while current code does not (for me at least).

Alphrag:
The added changes should fix #172.

vraravam (approved):
I don't use FB messenger. Approving without testing

Issue #151, opened 2022-08-16
[Recipe Request]: Lemmy


xgqt:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Name of the service

Lemmy

### URL of the service

https://lemmy.ml/

### GitHub URL to a recipe that has already been created by someone else or you

None found :(

### Features you want implemented

Creating a proper Lemmy service requires an ability to pick any instance.

https://join-lemmy.org/instances
Example Lemmy instances:
- https://lemmy.ml/
- https://feddit.it/
- https://szmer.info/

PR #154, opened 2022-08-17
recipes: add initial Lemmy service

xgqt:
Closes: https://github.com/ferdium/ferdium-recipes/issues/151
Signed-off-by: Maciej Barć <xgqt@gentoo.org>


xgqt:
@vraravam ping
Could you check it one out? Thanks in advance. :D

recipes/lemmy/webview.js
```diff
@@ -0,0 +1 @@
+module.exports = Ferdium => Ferdium;
```
vraravam, line 1: With such an empty webview.js, you will not get any badges. Is that ok?

vraravam (changes requested):

recipes/lemmy/webview.js
xgqt: If this is absolutely required I will ping You after I write code to support notification badges.
I'm not a JS programmer and this merge request was just meant to kick-start Lemmy support.

vraravam: Sure. But, conversely, as soon as this "shell" is introduced, we usually get folks filing bugs that this feature is not working (lol). I was trying to pre-empt that deluge. If you have any buddies who can contribute for this service (I dont use it, and so I cannot help without a login) - that would be great. If not, I can still approve it, but the bugs will have to stay unsupported till one of the contributors comes with their login to lemmy. I hope you understand that.

vraravam: If the above is ok, I can approve the PR

vraravam (approved):
Approving as-is. Haven't tested since I do not want to sign up for another service.

PR #155, opened 2022-08-18
add new recipe: GamingOnLinux

fschaupp:
add webview scraping message & notification count


vraravam (changes requested):
Where are the other files that are requires for the recipe to work?

fschaupp:
I just added the webview to the existing recipe.
Is such an update usually done in a new PR or somehow on the old, closed one?

vraravam:
Did you run the `pnpm run package` task? Please see the relevant documentation.

vraravam:
Whenever any recipe changes, at a minimum, the recipe's `package.json` needs to change the `version` value.

vraravam (changes requested):
When changing an existing recipe - you will need to ensure that the corresponding `package.json` file has the version bump. Please add that into this PR before it can be approved

Also, please squash all the commits in this PR branch and rebase from the latest master branch before this PR can be reviewed once more.

PR #174, opened 2022-09-12
Add service OnMail

Bass-03:
OnMail is a new [email service by Edison](https://www.onmail.com/).

Added unread count for all inboxes

### DIFF
--- a/recipes/onmail/package.json
+++ b/recipes/onmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "onmail",
   "name": "onMail",
-  "version": "1.0.2",
+  "version": "1.0.0",
   "license": "MIT",
   "config": {
     "serviceURL": "https://mail.onmail.com"
--- a/recipes/onmail/package.json
+++ b/recipes/onmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "onmail",
   "name": "onMail",
-  "version": "1.0.1",
+  "version": "1.0.2",
   "license": "MIT",
   "config": {
     "serviceURL": "https://mail.onmail.com"
--- a/recipes/onmail/package.json
+++ b/recipes/onmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "onmail",
   "name": "onMail",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "license": "MIT",
   "config": {
     "serviceURL": "https://mail.onmail.com"
--- a/recipes/onmail/package.json
+++ b/recipes/onmail/package.json
@@ -1,0 +1,9 @@
+{
+  "id": "onmail",
+  "name": "onMail",
+  "version": "1.0.0",
+  "license": "MIT",
+  "config": {
+    "serviceURL": "https://mail.onmail.com"
+  }
+}
--- a/recipes/onmail/webview.js
+++ b/recipes/onmail/webview.js
@@ -5,7 +5,6 @@ function _interopRequireDefault(obj) {
 }
 
 module.exports = Ferdium => {
-  // TODO: If your onMail service has unread messages, uncomment these lines to implement the logic for updating the badges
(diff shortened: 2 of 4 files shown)


recipes/onmail/webview.js
```diff
@@ -0,0 +1,23 @@
+const _path = _interopRequireDefault(require('path'));
+
+function _interopRequireDefault(obj) {
+  return obj && obj.__esModule ? obj : { default: obj };
+}
+
+module.exports = Ferdium => {
+  // TODO: If your onMail service has unread messages, uncomment these lines to implement the logic for updating the badges
```
vraravam, line 8: since you have implemented the logic, this `TODO` can be removed imo

recipes/onmail/webview.js
```diff
@@ -0,0 +1,23 @@
...
+  const getMessages = () => {
+    let countImportant = 0;
+    const inboxLinks = document.querySelectorAll('p.truncate');
+    for (const label of inboxLinks){
+      if (label.textContent) {
+        countImportant = label.nextSibling.textContent;
```
vraravam, line 14: Please see the other recipes. It would be safe to use `Ferdium.safeParseInt()` here

vraravam (changes requested):

recipes/onmail/webview.js
Bass-03: You are right, removing ...

recipes/onmail/webview.js
Bass-03: Also right, I didn't think it was needed, adding it...

Bass-03:
hey @vraravam changes done. 
Thanks!

recipes/onmail/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "id": "onmail",
+  "name": "onMail",
+  "version": "1.0.2",
```
vraravam, line 4: Please force commit this back to `1.0.0` since this is still in the PR

recipes/onmail/webview.js
```diff
@@ -0,0 +1,22 @@
...
+function _interopRequireDefault(obj) {
+  return obj && obj.__esModule ? obj : { default: obj };
+}
+
+module.exports = Ferdium => {
+  const getMessages = () => {
+    let countImportant = 0;
+    const inboxLinks = document.querySelectorAll('p.truncate');
+    for (const label of inboxLinks){
+      if (label.textContent) {
```
vraravam, line 12: In the if condition, you are checking for `label.textContent`, but the next line, you are using `label.nextSibling.textContent` - is that correct?

vraravam (changes requested):

recipes/onmail/webview.js
Bass-03: yes it is, the p is a sibling of the actual version, the p.truncate are a lot of labels, the only one we care about is the `inbox` one, so instead of defaulting to the first node, I decided to loop them and check, I think that is a better solution, in case the order changes or they add a new node before or what not.

recipes/onmail/package.json
Bass-03: I wanted to, but pnpm package is forcing me to have a version bump, Is there a way to avoid that?

recipes/onmail/webview.js
vraravam: since i dont have this service login, was just making sure that this was not an oversight. Thanks

recipes/onmail/package.json
vraravam: please fix this point alone - and we can approve the PR.

vraravam: `git commit --noverify`

vraravam (approved):
I don't use this service and so haven't tested this locally.

vraravam:
@all-contributors please add @Bass-03 for code

PR #175, opened 2022-09-12
Pnpx removal

Bass-03:
removed pnpx from scripts, that is deprecated, `pnpm exec` is needed, `pnpm <command>` is sufficient if there are no path conflicts, line in this case  

- Allow pnpm from version 7.6.0 and up
- tested with 7.10, works fine
- tested with 7.11, works fine

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,15 +5,15 @@
     "doc": "docs"
   },
   "engines": {
-    "pnpm": "7.6.0"
+    "pnpm": ">=7.6.0"
   },
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "prepare": "pnpm dlx husky install",
     "package": "node scripts/package.js",
     "create": "node scripts/create.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -5,15 +5,15 @@
     "doc": "docs"
   },
   "engines": {
-    "pnpm": "7.6.0"
+    "pnpm": ">=7.6.0"
```
vraravam, line 8: we don't really want to leave anything loose/open-ended. In the past, this has caused differences between developers' machines, which then become difficult to track down and fix.

vraravam: If you want to upgrade to the latest `pnpm` - please do that in the same PR and rephrase your commit messages/PR title.

vraravam (changes requested):

package.json
Bass-03: I have my local version of this file with no `pnpx` and allows my version of pnpm, so I am fine with that.

I think having to do a PR every time there is a `pnpm` release is overkill,  specially since they seem to release frequently, but suit yourselves :) 

I'll close this PR

PR #178, opened 2022-09-13
Fix: Unread count stays at 1 when all emails are read

Bass-03:
- Element that contains the unread count dissapears when count is zero.
- Added that scenario to reset th ecount to zero and send it to the badge

### DIFF
--- a/recipes/onmail/package.json
+++ b/recipes/onmail/package.json
@@ -1,7 +1,7 @@
 {
   "id": "onmail",
   "name": "onMail",
-  "version": "1.0.0",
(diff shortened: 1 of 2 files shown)


vraravam (changes requested):
Sorry, I thought I had already commented, but couldn't find it.

Please bump the version number in the `package.json` of this recipe without which we cannot approve this PR

vraravam (approved):

vraravam:
@all-contributors please add @Bass-03 for code

PR #207, opened 2022-10-18
Add google groups recipes

vijay94:
Added recipe for Google groups.


recipes/google-groups/webview.js
```diff
@@ -0,0 +1,22 @@
+module.exports = Ferdium => {
+  // if the user is on gmail's landing page, go to the login page.
+
+  const getMessages = () => {
+    let countImportant = 0;
+    console.info(countImportant);
+    const conversationLinks = document.querySelectorAll('div[aria-label="Conversations"]');
```
vraravam, line 7: This will not work for non-English locales.

vraravam (changes requested):

recipes/google-groups/webview.js
vijay94: @vraravam Made the changes and pushed.

PR #221, opened 2022-11-01
added ntfy support

jkischel:


jkischel:
hmmm. I don't quite understand what went wrong? 🤔

recipes/ntfy/webview.js
```diff
@@ -0,0 +1,14 @@
+module.exports = (Ferdium) => {
+  const getMessages = () => {
+    let msgRaw = document.title.match(/\(\d*\)/);
+    let messages = 0;
+
+    if (msgRaw) {
+      messages = parseInt(msgRaw[0].substring(1));
```
vraravam, line 7: Please use `Ferdium.safeParseInt` (you can take a look at how its used in one of the other recipes)

vraravam (changes requested):

PR #226, opened 2022-11-09
Add new recipe: PocketCasts

apraditya:
Add Pocket Casts recipe


vraravam (approved):

vraravam (commented):
Isn't this PR missing the webview.js? I'm not sure it'll work without it

apraditya:
I'm not sure the `webview.js` is required. I copied from `googlepodcasts` folder and update accordingly. I tried it by putting the folder in `~/Library/Application Support/Ferdium/recipes/dev`.

Anyways, what's your advise on this?

vraravam:
Please create the recipe using the commands present in the [documentation](https://github.com/ferdium/ferdium-recipes/blob/main/docs/integration.md#create-a-recipe). There's support for some files (eg `service.css`) which the user can directly customize. Without the `webview.js` in place, such customizations would not work, and then we will get bugs raised.
Also, if you based this off of the `googlepodcasts` recipe - please raise a PR to fix that as well. Thanks for pointing that out. I might have missed this point when reviewing and accepting that recipe submission.

vraravam (approved):

PR #235, opened 2022-11-11
fix: whatsapp message count selector

### LINKED ISSUE #230: [Bug]: WhatsApp wrong message count
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Ferdium Version

6.2.1-nightly.39

### Recipe Name

whatsapp

### Recipe Version

3.4.6

_No response_

### Steps to reproduce

Just create a WhatsApp service and use it I guess (I think it's due to updates to the website and not specific to my setup)

### Expected Behavior

The message count shows the correct count which is in my case 0 messages

### Actual Behavior

The message count shows 88 messages

### Screenshots

image

### Additional Information

_No response_

nathanaelhoun:
Fixes #230.

WhatsApp seems to have added `aria-label` to all components in each conversation, when they only used it for message counts previously.
The new selector only select the message count for each conversation, and so the count is correct as previously.

### DIFF
--- a/recipes/whatsapp/package.json
+++ b/recipes/whatsapp/package.json
@@ -1,7 +1,7 @@
 {
   "id": "whatsapp",
   "name": "WhatsApp",
-  "version": "3.4.6",
+  "version": "3.4.7",
   "license": "MIT",
   "config": {
     "serviceURL": "https://web.whatsapp.com",
--- a/recipes/whatsapp/webview.js
+++ b/recipes/whatsapp/webview.js
@@ -16,7 +16,7 @@ module.exports = Ferdium => {
       return;
     }
 
-    const unreadSpans = parentChatElem.querySelectorAll('span[aria-label]');
+    const unreadSpans = parentChatElem.querySelectorAll('div[role="gridcell"] > span > div > span[aria-label]');
     for (const unreadElem of unreadSpans) {
       const countValue = Ferdium.safeParseInt(unreadElem.textContent);
       if (countValue > 0) {


vraravam:
This still doesn't work for me. If anyone else can confirm, then we can merge this in

Alphrag:
It seems to work on my side both on the previous version on and the new one. But if that's not true for everyone, then we might need to find something else.

nathanaelhoun:
The selector is pretty fragile... We could try to find another logic to identify the unread counts badges.

cino:
I'll have a crack as well :)

cino:
The solution is actually working for me correctly, we could change the last part of the selector to only check for arialabels that include "unread" by changing to this: 

```
const unreadSpans = parentChatElem.querySelectorAll(
    'div[role="gridcell"] > span > div > span[aria-label*="unread"]',
);


But other than that it should be fine.

**Comment by vraravam** [comment]
> ```ts
> const unreadSpans = parentChatElem.querySelectorAll(
>     'div[role="gridcell"] > span > div > span[aria-label*="unread"]',
> );
> ```

Will this work for non-english languages?

**Comment by cino** [comment]
> > ```ts
> > const unreadSpans = parentChatElem.querySelectorAll(
> >     'div[role="gridcell"] > span > div > span[aria-label*="unread"]',
> > );
> > ```
> 
> Will this work for non-english languages?

Good one, ignore my comment.

**Comment by ryrun** [comment]
@vraravam 

Nope, not for the german version :/

Maybe we need to change the selector for ones with starting numbers, not sure for languages where non numbers are used, if there one:

`[ 
...parentChatElem.querySelectorAll('span[aria-label^="1"]'),
...parentChatElem.querySelectorAll('span[aria-label^="2"]'),
...parentChatElem.querySelectorAll('span[aria-label^="3"]'),
...parentChatElem.querySelectorAll('span[aria-label^="4"]'),
...parentChatElem.querySelectorAll('span[aria-label^="5"]'),
...parentChatElem.querySelectorAll('span[aria-label^="6"]'),
...parentChatElem.querySelectorAll('span[aria-label^="7"]'),
...parentChatElem.querySelectorAll('span[aria-label^="8"]'),
...parentChatElem.querySelectorAll('span[aria-label^="9"]'),
];`

**Comment by Zahrun** [comment]
This PR fixes the issue for me

**Comment by SpecialAro** [comment]
Just tried this PR and it doesn't work for me:
image

**Comment by victorbnl** [comment]
+1

**Comment by nathanaelhoun** [comment]
> @vraravam
> 
> Nope, not for the german version, atleast the version with "unread" selector.
> 
> Maybe we need to change the selector for ones with starting numbers, not sure for languages where non numbers are used, if there one:
> 
> ```js
> [ 
> ...parentChatElem.querySelectorAll('span[aria-label^="1"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="2"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="3"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="4"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="5"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="6"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="7"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="8"]'),
> ...parentChatElem.querySelectorAll('span[aria-label^="9"]'),
> ];
> ```

If I analyse this correctly, the `aria-label`s for contact names or messages starting with number will also match this selector :'(

**Comment by ryrun** [comment]
@nathanaelhoun As far as i can see none of my contacts have a filled aria-label for contact name and last message. But yeah, its still not optimal :/

**Comment by vraravam** [comment]
> @nathanaelhoun As far as i can see none of my contacts have a filled aria-label for contact name and last message. But yeah, its still not optimal :/
> 
> @victorbnl We could fetch all span's by aria-label and filter them out by text content (RegExp for numbers) and if border-radius is used. Could be a more stable solution.

just a reminder: we will also have to handle muted conversations

**Comment by victorbnl** [comment]
> @victorbnl We could fetch all span's by aria-label and filter them out by text content (RegExp for numbers) and if border-radius is used. Could be a more stable solution.

Good idea indeed

**Comment by felixmosh** [comment]
Is it possible to use title count?
image

**Comment by bernikr** [comment]
@felixmosh while it would be possible to use this, it doesn't differentiate between normal and muted checks.

**Comment by felixmosh** [comment]
> @felixmosh while it would be possible to use this, it doesn't differentiate between normal and muted checks.

What is mutated checks? 😅
It is possible to inject to that page a [mutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) code that will watch the title textContent and notify Ferdium regarding the change.

**Comment by bernikr** [comment]
I meant to write "chats". If you mute some chats in whatsapp their count will still be reported to ferdium, but it will be in a gray bubble instead of a red one. So it is necessary to not only extract the total unread count from the web page but also the unread count of muted chats.

**Comment by felixmosh** [comment]
> I meant to write "chats". If you mute some chats in whatsapp their count will still be reported to ferdium, but it will be in a gray bubble instead of a red one. So it is necessary to not only extract the total unread count from the web page but also the unread count of muted chats.

Ok, got it... Maybe it is possible to connect to the react app by simulating react-dev-tools ;)

**Comment by victorbnl** [comment]
> Ok, got it... Maybe it is possible to connect to the react app by simulating react-dev-tools ;)

It doesn't seem to be possible on production websites unfortunately

**Comment by victorbnl** [comment]
I think I just came up with the cleanest solution so far.

So basically I just discovered React stores properties in the DOM elements, under the __reactFiber$hash attribute. One of these properties is the key, which is a bit I think like the ID of the element, and gives us its usage. The key of the unread badge is `.$icon-unread` and the key of the muted one is `.$icon-muted`.

So here is my solution to get the elements (my function just returns a list of elements, I let you take my code and adapt it as you wish to generate a counter etc):

```js
function getElementsByKey(wantedKey) {
    let results = []

    // Walk the tree
    for (element of document.querySelectorAll("*")) {

        // Find the __reactFiber attribute name (for it has a hash in it)
        let fiberAttrName
        for (attr of Object.keys(element)) {
            if (attr.startsWith("__reactFiber")) {
                fiberAttrName = attr
                break
            }
        }

        // Get the current element's key
        if (fiberAttrName) {
            let elementKey = element[fiberAttrName].key

            // Check if the element's key is the one we want
            if (elementKey == wantedKey) {
                results.push(element)
            }
        }

    }

    return results
}

function test() {

    let start = performance.now()

    console.log("Unread badges:")
    console.log(getElementsByKey(".$icon-unread"))

    console.log("Muted badges:")
    console.log(getElementsByKey(".$icon-muted"))

    let end = performance.now()

    console.log(`Got these in ${end - start} ms`)
}

test()
```
*To test it, just copy paste the code in the devtools console*

I thought looping through litterally all the elements of the page would take a really long time but actually it takes only 9ms on my side, with a conversation open

felixmosh:
It is great idea! I've tried it and it works...

victorbnl:
I'm trying to actually implement it but I can't, for the `__reactFiber` attribute of DOM elements is not accessible from `webview.js` isolated context. I don't know if there is a way I can bypass this context isolation to get the attribute, with `webview-unsafe.js` maybe?

RaphaelJenni:
> Is it possible to use title count? image

Would it be possible to take this as a temporary fix? Having only the unmuted shown is better than having 123 messages.

SpecialAro:
> > Is it possible to use title count? image
> 
> Would it be possible to take this as a temporary fix? Having only the unmuted shown is better than having 123 messages.

I do like this approach and this is exactly what I've implemented locally. If you want to try it change all your whatsapp webview recipe file to:
```
const _path = _interopRequireDefault(require("path"));

function _interopRequireDefault(obj) {
  return obj && obj.__esModule ? obj : { default: obj };
}

module.exports = (Ferdium) => {
  const getMessages = () => {
    let count = 0;
    let indirectCount = 0;

    const title = document.querySelectorAll("title")[0].innerText;
    count = Ferdium.safeParseInt(
      title.substring(title.indexOf("(") + 1, title.indexOf(")"))
    );

    Ferdium.setBadge(count, indirectCount);
  };

  // inject webview hacking script
  Ferdium.injectJSUnsafe(_path.default.join(__dirname, "webview-unsafe.js"));

  const getActiveDialogTitle = () => {
    const element = document.querySelector("header .emoji-texttt");

    Ferdium.setDialogTitle(element ? element.textContent : "");
  };

  const loopFunc = () => {
    getMessages();
    getActiveDialogTitle();
  };

  window.addEventListener("beforeunload", async () => {
    Ferdium.releaseServiceWorkers();
  });

  Ferdium.handleDarkMode((isEnabled) => {
    if (isEnabled) {
      document.body.classList.add("dark");
    } else {
      document.body.classList.remove("dark");
    }
  });

  Ferdium.loop(loopFunc);

  Ferdium.injectCSS(_path.default.join(__dirname, "service.css"));
};
```

RaphaelJenni:
@SpecialAro Thanks for the snipped. It inspired me to checkout an other option.

Instead of accessing the HTML elements, why not directly access the indexedDB. I added a small snipped which works great:

```javascript
let unreadCount = 0;
        let unreadMutedCount = 0;

        const request = window.indexedDB.open("model-storage");
        request.onsuccess = (event) => {
            const db = request.result;
            const txn = db.transaction('chat', 'readonly');
            const store = txn.objectStore('chat');

            const query = store.getAll()

            query.onsuccess = (event) => {
                for (const chat of event.target.result) {
                    if (chat.unreadCount > 0) {
                        if (chat.muteExpiration > 0) {
                            unreadMutedCount += chat.unreadCount
                        } else {
                            unreadCount += chat.unreadCount
                        }
                    }
                }

                Ferdium.setBadge(unreadCount, unreadMutedCount);
            };
        };
```


The full webview.js looks as follows:

```javascript
const _path = _interopRequireDefault(require("path"));

function _interopRequireDefault(obj) {
    return obj && obj.__esModule ? obj : {default: obj};
}

module.exports = (Ferdium) => {
    const getMessages = () => {
        let unreadCount = 0;
        let unreadMutedCount = 0;

        const request = window.indexedDB.open("model-storage");
        request.onsuccess = (event) => {
            const db = request.result;
            const txn = db.transaction('chat', 'readonly');
            const store = txn.objectStore('chat');

            const query = store.getAll()

            query.onsuccess = (event) => {
                for (const chat of event.target.result) {
                    if (chat.unreadCount > 0) {
                        if (chat.muteExpiration > 0) {
                            unreadMutedCount += chat.unreadCount
                        } else {
                            unreadCount += chat.unreadCount
                        }
                    }
                }

                Ferdium.setBadge(unreadCount, unreadMutedCount);
            };
        };
    };

    // inject webview hacking script
    Ferdium.injectJSUnsafe(_path.default.join(__dirname, "webview-unsafe.js"));

    const getActiveDialogTitle = () => {
        const element = document.querySelector("header .emoji-texttt");

        Ferdium.setDialogTitle(element ? element.textContent : "");
    };

    const loopFunc = () => {
        getMessages();
        getActiveDialogTitle();
    };

    window.addEventListener("beforeunload", async () => {
        Ferdium.releaseServiceWorkers();
    });

    Ferdium.handleDarkMode((isEnabled) => {
        if (isEnabled) {
            document.body.classList.add("dark");
        } else {
            document.body.classList.remove("dark");
        }
    });

    Ferdium.loop(loopFunc);

    Ferdium.injectCSS(_path.default.join(__dirname, "service.css"));
};

```

Alphrag:
> Instead of accessing the HTML elements, why not directly access the indexedDB. I added a small snipped which works great:

Oh that could be a very good idea! It seems to works fine for me too, even though I don't understand how, since in the developer tools console, I can't debug it in the loop (never happens) and the values of `unreadCount` and `unreadMutedCount` are always 0. :face_with_spiral_eyes:

vraravam:
I still dont have the unread badge on the left sidebar. If this works for others, please go ahead and approve and squash + merge the PR.

SpecialAro:
@RaphaelJenni your solution works great and I think that is the cleanest one so far :D Would you be able to raise a PR request so that we can close this one as unmerged and accept yours ? Thank you very much!

nathanaelhoun:
Closing in favor of @RaphaelJenni's solution

RaphaelJenni:
Add new PR with discussed solution: https://github.com/ferdium/ferdium-recipes/pull/240

PR #236, opened 2022-11-12
ci: add recipe version bump check in PRs

nathanaelhoun:
As stated in the contributing documentation, recipe package version should be upgraded when it is updated. Currently, this change has to be checked by a maintainer when reviewing the PR, and it could be forgotten.

I wanted to play a bit with GitHub Actions, so I wrote a custom GA to check this version bump. Example of failing and passing commits can be found in https://github.com/nathanaelhoun/ferdium-recipes/pull/1.


nathanaelhoun:
Note: I am thinking about new recipes. I think the script is broken in this case. I have to test this case.

nathanaelhoun:
Updated the workflow file to not fail when adding a new recipe. This is ready!

SpecialAro (approved):
Haven't tested this myself but the logic seems correct! Thank you @nathanaelhoun 😄

PR #237, opened 2022-11-13
Fixed bug #230: whatsapp count

scheriam:
I tried to fix bug #230 by first finding out the proper CSS selector (here: class name) for the span with the unread count texts and then add the counts in its content.
Hope that this method is more stable as it relies on a variable name for the color of unread count badge.
Could not figure out what the previously contained height comparison is used here, so I left it out, but even if it is necessary, the first part of my code (to find out the CSS class name) could still be useful, as the class names are mini- and uglyfied.

### DIFF
--- a/recipes/whatsapp/package.json
+++ b/recipes/whatsapp/package.json
@@ -1,7 +1,7 @@
 {
   "id": "whatsapp",
   "name": "WhatsApp",
-  "version": "3.4.6",
+  "version": "3.4.7",
   "license": "MIT",
   "config": {
     "serviceURL": "https://web.whatsapp.com",
--- a/recipes/whatsapp/webview.js
+++ b/recipes/whatsapp/webview.js
@@ -8,10 +8,11 @@ module.exports = Ferdium => {
   const getMessages = () => {
   
     // first find the css selector for the unread count text
+    var unreadSelector;
     [...document.styleSheets].every(sheet => {
-      matchedRule = Array.from(sheet.cssRules).find(
+      const matchedRule = [...sheet.cssRules].find(
           rule => (
-            rule.type === CSSRule.STYLE_RULE && 
+            rule.constructor.name === "CSSStyleRule" && 
             rule.style['color'] === 'var(--unread-marker-text)'
           )
       );
@@ -21,9 +22,8 @@ module.exports = Ferdium => {
     let count = 0;
     let indirectCount = 0;
 
-    [...document.querySelectorAll(unreadSelector)].forEach(
-      unreadElem => {
-        const countValue = Ferdium.safeParseInt(unreadElem.innerText);
+    for (const unreadElem of document.querySelectorAll(unreadSelector)) {
+        const countValue = Ferdium.safeParseInt(unreadElem.textContent);
         if (
           !unreadElem.parentNode.previousSibling ||
           unreadElem.parentNode.previousSibling.querySelector(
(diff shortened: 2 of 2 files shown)


ruippeixotog:
Just FYI, there's an alternative method to fix this in https://github.com/getferdi/recipes/pull/894.

scheriam:
I am well aware of the other fix (well - I wasn't when I started mine... ;-) but after comparing both, although the other fix changes less code, I think that mine is a bit more stable in case of future changes of the DOM structure.

SpecialAro:
Hello! :D Just tried this fix and it didn't worked for me. I image that neither for @vraravam will work
image

SpecialAro:
> Hello! :D Just tried this fix and it didn't worked for me. I image that neither for @vraravam will work image

Hey again! Just tried this fix without dev mode and it working properly for me 😄 seems like in DEV mode the `Ferdium.setBadge` is not setting properly the badge count.

So I vote to merge this PR if @vraravam can also confirm it to be working on his machine!

oliverrahner:
This works for me (EN and DE languages).
Just two bugs:

1. Pinned and Muted

When you have a pinned **and** muted chat, it doesn't detect the muted state, because the pinned state is not the previous sibling then, but the sibling before:
image

image

Using the `if` clause fixes it for me:

```
        if (
          !unreadElem.parentNode.previousSibling ||
          unreadElem.parentNode.parentNode.querySelector(
            '[data-icon=muted]'
          ) === null
```

2. Manually marked unread

If a chat has been manually marked as unread, the result becomes NaN, because the badges doesn't contain a number:
image

Proposed fix:
````
        let countValue = parseInt(unreadElem.textContent);
        if (isNaN(countValue)) countValue = 1;
````

scheriam:
@oliverrahner thanks for the hint with the combined muted/pinned state.

I cannot reproduce the NaN situation, as I use `Ferdium.safeParseInt(unreadElem.textContent)` I could never trigger a `NaN` result, even with an empty badge.

But: it does not count towards the unread messages in that situation - did you mean that?

vraravam:
This PR still doesn't work for me. But, if it does work for a portion of the users, then at least that number of users won't keep raising duplicate bugs!

Screenshot 2022-11-17 at 6 06 40 AM

scheriam:
@vraravam I see from your screenshot that you tested with a group chat, is that correct? I just tested with direct conversations, so maybe that's the problem with the fix...

vraravam:
> @vraravam I see from your screenshot that you tested with a group chat, is that correct? I just tested with direct conversations, so maybe that's the problem with the fix...

yes - most of my chats are group chats. The expectation is that the solution should work for both

scheriam:
@vraravam I tested my fix now also successfully with a group chat. (It was also my expectation that the fix works with both ;-) Will continue testing a bit, but that will take some time.

oliverrahner:
> @oliverrahner thanks for the hint with the combined muted/pinned state.
> 
> ... as I use `Ferdium.safeParseInt(unreadElem.textContent)` I could never trigger a `NaN` result
> 
> But: it does not count towards the unread messages in that situation - did you mean that?

Oops, I tried the fix out in native Chrome, and thus hat to replace `Ferdium.safeParseInt()` by `parseInt()`, my bad...
But then, yes, if I mark a chat unread by intention I would at least like to be reminded of this in the unread count.

SpecialAro:
Closed in favor of: https://github.com/ferdium/ferdium-recipes/pull/235#issuecomment-1320027760 . A new PR should be open with this fix.

Thank you!

PR #247, opened 2022-11-24
fix whatsapp initial loading issue

balajiv113:


balajiv113:
Since we are relying on this database for count. It has impact on logout. Logout throws same blocked error

balajiv113:
> Since we are relying on this database for count. It has impact on logout. Logout throws same blocked error

Fixed this as well

vraravam (approved):

PR #255, opened 2022-12-01
Fix badge numbers for Any.do

chappy84:


vraravam:
@all-contributors please add @chappy84 for code.

PR #257, opened 2022-12-06
Get Element messages count from the spaces navbar

eandersons:
Element recipe was updated to get direct messages count from the spaces navigation bar as previously it was retrieved only from the active space resulting in situations when there are new messages in Element but there is no indicator if the currently selected Element space does not have unread notifications.

I am not completely sure if these changes do not break something, but from my side everything seemed fine.
Additionally, I could not find any setting in Element that would disable spaces, so I am considering these changes quite safe.

### DIFF
--- a/recipes/element/package.json
+++ b/recipes/element/package.json
@@ -1,7 +1,7 @@
 {
   "id": "element",
   "name": "Element",
-  "version": "1.3.0",
+  "version": "1.3.1",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


vraravam (changes requested):
Please fix the build breakages before we can move forward with this PR

vraravam:
@all-contributors please add @eandersons for code

Issue #268, opened 2022-12-21
[Bug]: Issues with the notifications count badge for Element


eandersons:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Ferdium Version

6.2.2

### Recipe Name

element

### Recipe Version

1.3.1

### Steps to reproduce

There are two cases:

1. built-in spaces "Favourites", "People" and "Rooms outside of a space" are enabled:
   1. Element settings should be opened;
   2. "Sidebar" should be selected;
   3. either of the built-in spaces ("Favorites", "People" and "Rooms outside of a space") or all of them should be enabled;
   4. wait until there are unread messages badge for at least one of those additional spaces;
2. unread messages count is greater than 1000:
   1. one has to be joined to a room that has a lot of activity;
   2. notifications has to be enabled for this room;
   3. wait until unread messages count reaches over 1000;

### Expected Behavior

There could be two scenarios for the expected behavior:

1. service notifications badge for Element has correct count (or at least as close as possible);
  this could be achieved by:

    - not adding notifications count from the additional built-in spaces;
    - multiplying floats from combined numbers (1.2K -> 1.2) by 1000 and then adding 50 (in this case resulting in 1250);
      this would ensure that difference from the actual count would not be greater than 50 as with such amount of unread messages it would not matter much anyway;

2. Ferdium would do the same as Element - display unread rooms count;
   this value can be easily extracted from `<title>`.

### Actual Behavior

1. Notifications badge for the default space ("Home") already has combined count of unread messages (including those from "Favorites", "People" and "Rooms outside of a space"), therefore service badge has greater value that actual notifications count.
2. Combined counts (1.2K, 2.3K) are interpreted as integers (1, 2).

The added screenshot illustrates both cases together:

- 4.2K is interpreted as 4;
- count from the space "Rooms outside of a space" is also added to the direct messages count.

That results in incorrect unread messages count - service badge has the value 742, but it should be 738 (assuming that "Rooms outside of a space" actually had 4 notifications).

### Screenshots

[...]

PR #271, opened 2022-12-28
Fix unread notifications count for Element

eandersons:
The following fixes were made:

- take only the space "Home" and ignore other built-in spaces (if they are enabled) as count in "Home" badge includes notifications from other built-in spaces as well;
- calculate approximate count from composite numbers (1.2K, 2.3K).

This is one of the two suggested solutions for #268.
Personally I prefer #272 over this pull request.

Before:
Element before

After:
Element after - guessed count

If this pull request is accepted, #272 should be closed.

### DIFF
--- a/recipes/element/webview.js
+++ b/recipes/element/webview.js
@@ -2,9 +2,20 @@ module.exports = Ferdium => {
   function getMessages() {
     let directCount = 0;
     const spacesBar = document.querySelector('.mx_SpaceTreeLevel');
-    for (const badge of spacesBar.querySelectorAll('.mx_NotificationBadge_count')) {
-      directCount += Ferdium.safeParseInt(badge.textContent);
+
+    for (const badge of spacesBar.querySelectorAll(
+      '.mx_SpaceItem:not(.mx_SpaceItem_narrow):first-of-type .mx_NotificationBadge_count, '
+        + '.mx_SpaceItem_narrow .mx_NotificationBadge_count'
+    )) {
+      const badgeContent = badge.textContent.toLowerCase();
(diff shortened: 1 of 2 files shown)


recipes/element/webview.js
```diff
@@ -2,9 +2,20 @@ module.exports = Ferdium => {
...
-    for (const badge of spacesBar.querySelectorAll('.mx_NotificationBadge_count')) {
-      directCount += Ferdium.safeParseInt(badge.textContent);
+
+    for (const badge of spacesBar.querySelectorAll(
+      '.mx_SpaceItem:not(.mx_SpaceItem_narrow):first-of-type .mx_NotificationBadge_count, '
+        + '.mx_SpaceItem_narrow .mx_NotificationBadge_count'
+    )) {
+      const badgeContent = badge.textContent.toLowerCase();
+      directCount += badgeContent.endsWith('k')
+        ? Number.parseFloat(badgeContent) * 1000 + Number.parseInt(
```
vraravam, line 12: I dont understand the 2nd `Number.parseInt` that's being added after the multiplication by 1000. Can you please explain with an example?

vraravam (changes requested):

recipes/element/webview.js
eandersons: If the k-number integer has one digit, the closest approximation is 50: 1.2 * 1000 + 50 = 1250.

If the k-number integer has two digits, the closest approximation is 500: 12 * 1000 + 500 = 12500 (because 12500 is the closest number to both 12000 and 13000).

And so on.

vraravam:
Closing in favor of #272

PR #272, opened 2022-12-28
Use unread rooms count like Element app does

eandersons:
This change uses unread rooms instead of messages count like Element (web)app does
Unread rooms count is extracted from the `title` tag.

This is one of the two suggested solutions for #268.
Personally I prefer this over #271.

Before:
Element before

After:
Element after - unread rooms count

If this pull request is accepted, #271 should be closed.

### DIFF
--- a/recipes/element/package.json
+++ b/recipes/element/package.json
@@ -1,7 +1,7 @@
 {
   "id": "element",
   "name": "Element",
-  "version": "1.3.2",
+  "version": "1.3.3",
   "license": "MIT",
   "aliases": [
     "Riot.im",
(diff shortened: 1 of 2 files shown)


vraravam:
Since this manner of counting is different from how the web version of element works, I dont agree on this as being the counting method. I propose to close this without mering.

@SpecialAro @kris7t @muhamedsalih-tw @balajiv113 - wdyt?

eandersons:
@vraravam, no, Element WEB version is using unread rooms count in badge:
image

This behaviour is the same as in desktop app:
image

vraravam (approved):
Since this seems closer to how the native thick client works, approving this PR.

vraravam:
@all-contributors please add @eandersons for code

PR #273, opened 2022-12-31
Fix: Infomaniak unread messages badge includes spam emails

RaphaelJenni:
Until now, the infomaniak recipe showed a number of unread messages in the inbox.
If the inbox was empty it selected the first folder with unread messages, which was often the spam folder.
With this change, the unread messages badge now always only selects the first folder (the inbox) for displaying a number. Unread mails in other folders now only lead to a blue batch, but no number.


recipes/infomaniak-mail/webview.js
```diff
@@ -1,8 +1,25 @@
 module.exports = Ferdium => {
   const getMessages = () => {
-    const count = document.querySelector('.ws-tree-node-badge');
-    const countText = count ? count.textContent : null;
-    Ferdium.setBadge(count && countText ? Number(countText) : 0);
+    // This selects the first folder (the inbox and reads its unread messages count)
+    const inboxField = document.querySelector('.ws-tree-node-content')
+    const inboxCountField = inboxField.querySelector('.ws-tree-node-badge');
+    const inboxCountText = inboxCountField ? inboxCountField.textContent : null;
+    const inboxCount = inboxCountField && inboxCountText ? Number(inboxCountText) : 0;
```
vraravam, line 7: For the condition, `inboxCountField` is not necessary to be checked, since if that was undefined or null, then the prevopus line would have automatically returned a null. So, checking only for `inboxCountText` is enough here. Also, please use `Ferdium.safeParseInt` instead of the `Number` constructor.

recipes/infomaniak-mail/webview.js
```diff
@@ -1,8 +1,25 @@
...
+    
+    const importantBadge = inboxCount;
+    let unimportantBadge = 0;
+    
+    if (inboxCount == 0){
+      // This selects the first folder with an unread message count.
+      // The actaul count and the total of all other folders is not needed as the badge has no number.
+      const totalCountField = document.querySelector('.ws-tree-node-badge');
+      const totalCountText = totalCountField ? totalCountField.textContent : null;
+      const totalCount = totalCountField && totalCountText ? Number(totalCountText) : 0;
```
vraravam, line 17: Same changes here for the ternary condition as well as the Number parsing.

vraravam (changes requested):

RaphaelJenni:
@vraravam Thanks for the review. I didn't change those things in the first place, as they were already present in the existing version. The issues are now resolved.

vraravam (approved):

vraravam:
@all-contributors please add @RaphaelJenni for code

PR #290, opened 2023-01-16
feat: add darkmode to chatgpt recipe

SpecialAro:
Add darkmode handling to ChatGPT

### DIFF
--- a/recipes/chatgpt/webview.js
+++ b/recipes/chatgpt/webview.js
@@ -1,1 +1,9 @@
-module.exports = Ferdium => Ferdium;
+module.exports = Ferdium => {
+  Ferdium.handleDarkMode(isEnabled => {
+    if (isEnabled) {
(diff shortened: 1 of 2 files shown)


recipes/chatgpt/webview.js
```diff
@@ -1 +1,9 @@
-module.exports = Ferdium => Ferdium;
+module.exports = Ferdium => {
+  Ferdium.handleDarkMode(isEnabled => {
+    if (isEnabled) {
```
vraravam, line 3: Please collapse this `if` block into a ternary condition since its a single word difference - easier to maintain

vraravam (changes requested):

recipes/chatgpt/webview.js
SpecialAro: agree!

vraravam (approved):

Issue #284, opened 2023-01-11
[Recipe Request]: Instagram Direct (ie messages only)


adamlove86:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Name of the service

Instagram Direct

### URL of the service

https://www.instagram.com/direct/inbox/

### GitHub URL to a recipe that has already been created by someone else or you

https://github.com/miguelcolmenares/franz-recipe-instagram-direct-messages

### Features you want implemented

Have tried miguelcolmenares recipe but it doesn't seem to work when I clone into either ~/snap/ferdium/203/.config/Ferdium/recipes/dev or ~/.config/Ferdi/recipes/dev/
Given the popularity of instagram over, for example, facebook, it'd be great to have a messaging only feature - without the distracting feed!
I'm currently using 'custom webpage' https://www.instagram.com/direct/inbox/, but it doesn't refresh very well

Issue #280, opened 2023-01-05
[Recipe Request]: DEV.to


s-shubham-22:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Name of the service

DEV Community

### URL of the service

https://dev.to/

### Features you want implemented

I want DEV Community implemented in Ferdium Community so we don't have to go elsewhere for some of the services or web apps.

Issue #279, opened 2023-01-05
[Recipe Request]: Medium


s-shubham-22:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Name of the service

Medium

### URL of the service

https://medium.com/

### Features you want implemented

I want Medium website implemented in Ferdium

Issue #256, opened 2022-12-02
[Recipe Request]: HelpScout


mihha:
### Preflight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I have searched the [issue tracker](https://github.com/ferdium/ferdium-recipes/issues) for a feature request that matches the one I want to file, without success.

### Name of the service

HelpScout

### URL of the service

https://www.helpscout.com/

### GitHub URL to a recipe that has already been created by someone else or you

https://github.com/extendfranz/help-scout

### Features you want implemented

I have tried to install this recipe as described on the GitHub of the service, but after that my Ferdium crashed

Since I am not good with JavaScript I can't figure out what could be wrong

PR #293, opened 2023-01-19
feat: add several recipes

SpecialAro:
This PR is a batch of several new recipes. I made this so we can cleanup some of the requests on the issue board.
Some logic is still needed in order to close some PRs (as I indicate bellow), nevertheless, we can merge as it is and implement it afterwards.

Step forward for https://github.com/ferdium/ferdium-recipes/issues/264 (missing notification)
Step forward for https://github.com/ferdium/ferdium-recipes/issues/183
Step forward for https://github.com/ferdium/ferdium-recipes/issues/227
Step forward for https://github.com/ferdium/ferdium-recipes/issues/281 although I think it won't be possible to implement what the user asks, for the time being.
Step forward for https://github.com/ferdium/ferdium-recipes/issues/249 - not sure how to implement what the user requests.
Step forward for https://github.com/ferdium/ferdium-recipes/issues/286
Step forward for https://github.com/ferdium/ferdium-recipes/issues/121
Closes https://github.com/ferdium/ferdium-recipes/issues/280
Closes https://github.com/ferdium/ferdium-recipes/issues/191
Closes https://github.com/ferdium/ferdium-recipes/issues/209
Closes https://github.com/ferdium/ferdium-recipes/issues/254
Closes https://github.com/ferdium/ferdium-recipes/issues/256
Closes https://github.com/ferdium/ferdium-recipes/issues/244
Closes https://github.com/ferdium/ferdium-recipes/issues/284
Closes https://github.com/ferdium/ferdium-recipes/issues/279
Closes https://github.com/ferdium/ferdium-recipes/issues/204
Closes https://github.com/ferdium/ferdium-recipes/issues/130
Closes https://github.com/ferdium/ferdium-recipes/issues/206

image


SpecialAro:
> If you take a look at https://github.com/ferdium/ferdium-recipes/blob/main/scripts/sample_recipe/webview.js - the auto-injection of the `service.css` is a semi-hidden feature for all recipes - and this is available for the user to create such a file from the Settings pane. If the user then creates such a file, the missing lines in most of these `webview.js` would mean that the loading of those user-created files will not work.

So, just to clarify, we should include that code in every recipe created in this PR?

SpecialAro:
> > So, just to clarify, we should include that code in every recipe created in this PR?
> 
> yes, unfortunately.

Maybe we should make a PR to fix this so it is consistent for every recipe in the future.

vraravam:
> Maybe we should make a PR to fix this so it is consistent for every recipe in the future.

The last time I checked, this was generated correctly when running the create script to generate the shell for any new recipe. Isn't that what you observed for this PR?

vraravam (approved):

PR #308, opened 2023-02-03
Fix unread notifications count for yahoo-mail

dobaniashish:
`data-test-unread-count` attribute does not exist anymore. Using text content of notification `span` element.


vraravam (changes requested):
Please read the docs and update the version in the package.json.
If you had run the `pnpm package` target, it should have errored out

vraravam (approved):

vraravam:
@all-contributors please add @dobaniashish for code

PR #333, opened 2023-04-21
Add DeepL recipe

victorbnl:
#### Pre-flight Checklist

#### Description of Change

I added a DeepL recipe, closes https://github.com/ferdium/ferdium-app/issues/1103


Alphrag:
Thanks @victorbnl I have added the code to support custom css as it is better to always include it in case a user fancies adding one.

@all-contributors please add @victorbnl for code

victorbnl:
> Thanks @victorbnl I have added the code to support custom css as it is better to always include it in case a user fancies adding one.

I don’t see the point as the file you’re mentioning is `service.css`, not `user.css`. And to my knowledge, the user deals with user.css, not service.css, as service.css is for the recipe developer. So I didn’t add the code as I had no need for a service.css file. But if it’s better to add it everytime then let’s do it.

PR #343, opened 2023-04-27
Add Gather service

victorbnl:
#### Pre-flight Checklist

#### Description of Change

I added the [Gather](https://www.gather.town/) service, closes https://github.com/ferdium/ferdium-app/issues/1132.


victorbnl:
Is it normal that the version bump check fails even though I added a new service?

Alphrag:
It shouldn't be the case, but there seems to be a problem when we add a new service with the action we're using that only checks against the last commit instead of the diff between the PR branch and the head of the target where it can't find the file. 
I am not sure yet exactly where it is coming from (thinking of two different points) but this can be verified locally for now.

Alphrag:
It shouldn't be the case, but there seems to be a problem when we add a new service with the action we're using that only checks against the last commit instead of the diff between the PR branch and the head of the target where it can't find the file. 
I am not sure yet exactly where it is coming from (thinking of two different points) but this can be verified locally for now.

victorbnl:
Alright then!

PR #353, opened 2023-05-04
Update Office 365 Outlook service

Thelm76:
update domain from live.com to office.com

#### Pre-flight Checklist

- [ ] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [ ] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

### DIFF
--- a/recipes/office365-owa/package.json
+++ b/recipes/office365-owa/package.json
@@ -1,14 +1,15 @@
 {
   "id": "office365-owa",
   "name": "Office 365 Outlook",
-  "version": "1.6.2",
+  "version": "1.7.0",
   "license": "MIT",
   "aliases": [
+    "office.com"
     "live.com",
     "outlook"
   ],
   "config": {
-    "serviceURL": "https://outlook.live.com/mail/",
+    "serviceURL": "https://outlook.office.com/mail/",
     "hasNotificationSound": true,
     "hasHostedOption": true,
     "hasCustomUrl": true,


Thelm76:
Hi, I've not run any tests yet (just changed the URL in my local service)

The `Contributing guidelines` link is dead, so I assume I should look at [here](https://github.com/ferdium/ferdium-recipes/blob/main/docs/updating.md)

Also, I didn't see ant policy regarding version numbers, so I also assumed that changing the service url is a relatively major change, so I bumped to 1.7.0

Alphrag:
Hi @Thelm76 thanks for the PR. 
However, I don't think we will change that at the moment. Indeed, the usage of either urls `live.com` and `office.com` seem to be the norm. I don't know which one is globally more present than the other, but both are definitely very common. For this reason, we added the "self-hosting" setting for that recipe so that users could set the domain name that correspond to the account they connect to (I personally use both domains as the different emails I log in behave differently). Changing now which one is the default would disconnect some users which are currently staying connected.

Thelm76:
Oh I see... I hadn't tested with my former live address and yup' it doesn't work... Easiest setting for users would be to have a 3-buttons selection at setup (as it is not very clear that we have to switch from live.com to office.com)...

Thelm76:
I'll work on this when I get some time to dig in this

mcmxcdev:
Hey @Thelm76, any chance that you can finish up this PR? Otherwise, I will close it.

vraravam:
Closing due to inactivity

PR #358, opened 2023-05-09
Add Elk recipe

jalil-salame:
#### Pre-flight Checklist

#### Description of Change

Add a recipe for Elk.

It does support notifications, but I feel like the way I'm doing it is way too janky.

Elk is a modern Mastodon web client, you can check it out at https://elk.zone/


recipes/elk/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "elk",
+  "name": "Elk",
+  "version": "0.1.0",
```
vraravam, line 4: please change the version of the recipe to `1.0.0`

recipes/elk/webview.js
```diff
@@ -0,0 +1,9 @@
+module.exports = (Ferdium) => {
+    Ferdium.loop(() => {
+        // FIXME: Don't assume 0th element exists
```
vraravam, line 3: Please fix this with some defensive programming.

vraravam (changes requested):

recipes/elk/webview.js
```diff
@@ -0,0 +1,9 @@
...
+        const notifications = document.querySelectorAll("[href$=notifications]")[0].outerText;
+        let parsedValue = Ferdium.safeParseInt(notifications);
+
+        Ferdium.setBadge(parsedValue);
+    });
```
vraravam, line 8: There should be some lines that are used to load `service.css`. please refer to the template and add them.

vraravam (changes requested):

recipes/elk/webview.js
jalil-salame: Is it necessary? I don't have a service.css file

recipes/elk/package.json
jalil-salame: I changed it to 1.0.1 because the hook was complaining about changing the package without a version bump after I set it to 1.0.0

recipes/elk/webview.js
jalil-salame: I fixed this by testing that the array is not empty (length === 0), is this enough?

recipes/elk/webview.js
vraravam: That option is for the user to be able to override if they choose to - not for the recipe to have a service.css

recipes/elk/webview.js
vraravam: you have a `FIXME` comment. If you have identified any scenario where this can go wrong, please fix that, else remove the comment.

recipes/elk/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "id": "elk",
+  "name": "Elk",
+  "version": "1.0.1",
```
vraravam, line 4: you can override that check by committing the change with the `-n` switch for `git commit` command

recipes/elk/webview.js
```diff
@@ -0,0 +1,17 @@
...
+
+    // Get notifications
+    Ferdium.loop(() => {
+        const notifications = document.querySelectorAll("[href$=notifications]");
+        if (notifications.length === 0) {
+          return;
+        }
+
+        // Assume first element contains the number of notifications
+        let parsedValue = Ferdium.safeParseInt(notifications[0].outerText);
```
vraravam, line 13: Is it possible for `notifications[0]` to be `undefined` or `null`? If so, you will still need to do some defensive programming here imo

vraravam (changes requested):

jalil-salame:
The Check Recipe Version Bump action should be skipped in this case, right?

recipes/elk/webview.js
```diff
@@ -0,0 +1,19 @@
+module.exports = (Ferdium) => {
+  const path = require('path');
```
vraravam, line 2: Please use the correct code from the template. This will not work due to the security checks.

vraravam (changes requested):

vraravam (approved):

vraravam:
@all-contributors please add @jalil-salame for code

PR #359, opened 2023-05-10
Add MiroTalk to Ferdium recipes.

miroslavpejic85:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [ ] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change
Add [MiroTalk](https://www.reddit.com/r/selfhosted/comments/11kt8zi/free_selfhosted_webrtc_alternative_to_zoom_teams/?utm_source=share&utm_medium=web2x&context=3) to Ferdium recipes.

Thank you!

PR #360, opened 2023-05-11
Add Nextcloud Deck Recipe

mindcombatant:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [ ] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [✔️] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change

Added Nextcloud Deck Recipe using other Nextcloud Recipes as a base. Icon was edited by myself. See recipes/nextcloud-deck/README.md for further details.

PR #362, opened 2023-05-11
fix(ci): version bump check when new recipe and multiple commits

nathanaelhoun:
Fix the check of recipe version bump by the CI, which was failing when adding a new recipe with 2+ commits in the PR. (example: https://github.com/ferdium/ferdium-recipes/pull/359).

We are now using another gh-action for detected changed files, which seems to work ok.

Here are some tests:
image


nathanaelhoun:
Hi @Alphrag, if it seems go to you, you can merge it, I will be away for the week-end.

After the merge, rebasing #360 should resolve the failing CI.

.github/workflows/check-pr.yml
```diff
@@ -14,11 +14,11 @@ jobs:
         uses: actions/checkout@v3
       - name: Fetch main branch
         run: git fetch origin main --depth 1
-      - uses: lots0logs/gh-action-get-changed-files@2.1.4
+      - uses: tj-actions/changed-files@v35
         id: changed-files
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
```
Alphrag, line 20: The action `tj-actions/changed-files` does not appear to need the token (there is no input parameter of that kind), so this can be removed.

Alphrag (changes requested):

Alphrag:
This is exactly the action I thought of when I realised the previous one was not working as expected, thanks @nathanaelhoun for checking was changes were required and testing it.

Alphrag (approved):

PR #367, opened 2023-05-19
Add badge count support to AirMessage recipe

cloopadoop:
#### Pre-flight Checklist

#### Description of Change
- Updated service icon
- Cycle through applicable DOM elements and check for bold font weight, which indicates unread messages in a thread


cloopadoop:
(By the way, Contributing Guidelines file is gone!)

recipes/airmessage/webview.js
```diff
@@ -1,8 +1,15 @@
...
-  //   // TODO: Insert your notification-finding code here
-  //   Ferdium.setBadge(0, 0);
-  // };
-  // Ferdium.loop(getMessages);
+module.exports = (Ferdium) => {
+  const getMessages = function getMessages() {
+    let count = 0;
+    let messageNameElements = document.getElementsByClassName("MuiTypography-body1");
+    if (messageNameElements.length > 0) {
+      for (i = 0; i < messageNameElements.length; i++) {
```
Alphrag, line 6: Please fix this `for` loop. I would suggest making it more robust by using the automatic loops (in the like of `for (var element of messageNameElements)` which makes the test on the previous line unnecessary as well.

Alphrag (changes requested):
I am not using this service so I cannot verify if this will work appropriately. However, I think that relying on the font weight of elements is prone to break and I would recommend finding a more robust approach for the future. If this still works and you think that is at the moment the best way, I will be fine with it though.

On another note, please make sure to run the tests as [described there](https://github.com/ferdium/ferdium-recipes/blob/main/docs/updating.md#5-running-the-validation-checks) before you push the next changes, and this should be done before submitting a PR as well so that linting and errors are caught as early as possible.

PR #370, opened 2023-05-23
feat(tt-rss): Optionally limit notifications to a Category.

thehale:
#### Pre-flight Checklist

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change

As described in the updated README for tt-rss, users can now choose to limit notifications to only new articles published under a Category of their choosing. Articles published in other Categories are then considered "indirect" messages.

This feature is made available via parsing a custom URL parameter from the Service URL. Preferrably Ferdium would enable recipies to request custom fields from users, but for now this workaround accomplishes the task.

### DIFF
--- a/recipes/tt-rss/README.md
+++ b/recipes/tt-rss/README.md
@@ -1,3 +1,39 @@
 # Tiny Tiny RSS for Ferdium
 
-Support for [Tiny Tiny RSS](https://tt-rss.org/)
+Interact with [Tiny Tiny RSS](https://tt-rss.org/) inside of Ferdium!
+
+## Usage
+
+1. Search for "RSS" in Ferdium's **Available Services**
+2. Enter the URL of your Tiny Tiny RSS server
+3. Save the service and log in!
+
+## Configuration
+
+### `ferdiumNotificationLabel`
+
+You can limit notifications from Tiny Tiny RSS to new articles in a specific
+Category using the custom `ferdiumNotificationLabel` query parameter as part of
+your server URL.
+
+```
+https://ttrss.example.com/tt-rss/?ferdiumNotificationLabel=CATEGORY_NAME
+```
+
+For example, if I wanted to only see notifications from a Category titled
+`Software - Updates`, I would use one of the following query parameters:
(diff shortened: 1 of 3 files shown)


vraravam (approved):
I don't use this service, but the structure of the code submission looks ok. approving

Alphrag:
Same things for me, I do not use such a service but all seems good.
It is true that if recipes could add some parameters in their `json` configuration as a list that could be used in the service settings of the app, this could open new directions. If you know how to deal with it and fancy making a PR to make that happen, we will be happy to look at it 😉

Alphrag (approved):

thehale:
Here's a visual example of how this feature works:

Add the query parameter `?ferdiumNotificationCategory=CATEGORY_NAME` to the URL in the app settings. (e.g. "Software - Updates")

Then the icon will only show "direct message" notifications for new articles under that CATEGORY_NAME.
image

If that CATEGORY_NAME doesn't have any new articles, then the icon will show the "indirect messages" icon.
image

Omitting the query parameter will count all new articles as "direct messages" (preserving the current functionality).

thehale:
> Same things for me, I do not use such a service but all seems good. It is true that if recipes could add some parameters in their `json` configuration as a list that could be used in the service settings of the app, this could open new directions. If you know how to deal with it and fancy making a PR to make that happen, we will be happy to look at it 😉

Since these were my first contributions to Ferdium, I stayed on the simpler side. If my needs grow on this front, then I'll wrestle with Ferdium's core code itself to explore these possibilities. I'd imagine we'd need to support some sort of dynamic UI creation (for each custom setting/field, e.g. based on it's data type) and new recipe APIs for reading the data stored there.

Alas, this is probably better handled as a feature request on ferdium-app.

PR #393, opened 2023-07-19
Add Excalidraw recipe

victorbnl:
#### Pre-flight Checklist

#### Description of Change

Closes https://github.com/ferdium/ferdium-app/issues/1223


vraravam (changes requested):
I dont think we should have 2 recipes for Excalidraw. Other than the `serviceURL`, what else is different?

victorbnl:
Pretty much nothing for the recipe, but having only one recipe makes it harder for either free or paid users I think. Say we put as a default the URL of the free version, paid users will have to type in the URL manually.

vraravam:
> paid users will have to type in the URL manually.

you can add a self-hosted option in the recipe's package.json for them to then configure it

victorbnl:
They will still have to configure it manually.

Actually the best thing I’m thinking about is providing the package.json with an array of URLs and having Ferdium to ask which URL to use when configuring the service, but it would require changes in Ferdium as well.

victorbnl:
Well I’ll make the free version the default one and add a self-hosted option.

vraravam (approved):

PR #401, opened 2023-07-27
Wikipedia recipe

meiphoo7-Mae:
wikipedia recipe

#### Pre-flight Checklist

- [ ] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [ ] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change

This is a proposal for a new recipe for easy access to Wikipedia. It accesses the general Wikipedia site (wikipedia.org).

Notice: I created the recipe manually based upon existing recipes. I did not follow the guide outlined here: https://github.com/ferdium/ferdium-recipes/blob/main/docs/integration.md because it somehow does not work on my Linux computer and presents errors. However based on the examples I used I believe the recipe is correct. If it is not please let me know.


victorbnl:
I’d just name it “Wikipedia” rather than “Wikipedia The Free Encyclopedia”.

meiphoo7-Mae:
@victorbnl  “Wikipedia The Free Encyclopedia” is the official name of the project which is also included on any Wikipedia site. This is the same as in your Perplexity recipe. That site is usually referred to as "Perplexity". Your recipe names it (correctly) Perplexity AI.

victorbnl:
Perplexity is referred to as “Perplexity AI” at least in the tab title.

On Wikipedia’s home page, the tab title is “Wikipedia”. As for the logo, “The Free Encyclopedia”, by its style, looks to me more like a subtitle than a part of the name.

Also if you go to [Wikipedia’s “About” page](https://en.wikipedia.org/wiki/Wikipedia:About), it is referred to as “Wikipedia, the free encyclopedia” once and “Wikipedia” more than ten times. That makes me think that in the first occurence, the “the free encyclopedia” part was not part of the name but just an explanation of what it is.

However I would like to know where you saw it as the official name of the project.

meiphoo7-Mae:
I ran it through Perplexity, which tells that the official name is just "Wikipedia", also the official logo is the one that I've included in the recipe. So it remains a mystery why all wikipedia sites have the phrase "The Free Encyclopedia" (or the localized version of it) in the logo and apparently do not display the official logo.
Anyway, I've changed the recipe to display just "Wikipedia" and I've made it possible to add a custom url because I prefer the localized version of my country instead of the general site. Because there are 300+ localized sites hardcoding them in the recipe is not really an option.

victorbnl:
This version bump right from 1.0.0 to 1.2.2 looks strange and is useless but I don’t know if this is a problem. I don’t think it is technically, but let’s see what say the developers. Other than that it looks good to me.

meiphoo7-Mae:
Thanks, for your input.
Going from the previous version to the current version I've made (in my opinion) two larger changes and two smaller changes. Hence the version bump to 1.2.2. During my tests not everything worked as expected so some things are changed twice. I did not commit the intermediate versions.
I don't really care about the version number so I'll leave it to the developers to decide what it should be if the recipe is going to be accepted.

victorbnl:
The thing is your changes do not really matter because you are adding a recipe, and you made the changes locally. The version number is used so that clients know when to update the recipes. However here there is nothing to update, because nobody has the Wikipedia recipe yet, so you could as well have not increased the version number at all. Similarly when doing multiple local changes, it’s enough to increase the version number only once as it will update everyone’s recipe.

Nevertheless it is only my take on it, maybe developers will agree with increasing the version number on every local change and even for new recipes, so please correct me if I’m wrong.

recipes/wikipedia/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "id": "wikipedia",
+  "name": "Wikipedia",
+  "version": "1.2.2",
```
SpecialAro, line 4: Can we name it "1.0.0" as this is the first version of the Recipe? Thank you!

SpecialAro (changes requested):
Overall, LGTM! But please, fix my comment so I can approve and merge the recipe.

Thank you @meiphoo7-Mae !

meiphoo7-Mae:
@SpecialAro Per your request I've changed the version to 1.0.0

SpecialAro:
Thank you @meiphoo7-Mae !

meiphoo7-Mae:
@SpecialAro
Just checking...
Can I delete this (temporary) repository?

SpecialAro (approved):

SpecialAro:
My bad @meiphoo7-Mae. I thought I've approved and merged it already. Doing it now! Thank you!

SpecialAro (changes requested):
Just noticed that there is a problem with the build command, probably caused by lint. I'll fix it today if I get the change, so please don't delete the repo until this is merged @meiphoo7-Mae

meiphoo7-Mae:
Thanks, I'll keep the repository active.

SpecialAro (approved):

SpecialAro:
@allcontributors-bot - please add @meiphoo7-Mae for code

SpecialAro:
@all-contributors - please add @meiphoo7-Mae for code

PR #402, opened 2023-07-30
Update google-calendar recipe to load external scripts to fix CSP error

vraravam:
#### Pre-flight Checklist

#### Description of Change

Based on https://github.com/ferdium/ferdium-app/issues/1086\#issuecomment-1656926137, fixed google calendar recipe to load the external css.

### DIFF
--- a/recipes/google-calendar/package.json
+++ b/recipes/google-calendar/package.json
@@ -1,7 +1,7 @@
 {
   "id": "google-calendar",
   "name": "Google Calendar",
-  "version": "2.4.0",
+  "version": "2.4.1",
   "license": "MIT",
   "aliases": [
     "google-calendar",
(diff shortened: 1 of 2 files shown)


mcmxcdev:
Veriified locally that it works:

- went to `/recipes` folder inside `ferdium-app` and applied the same changes as here
- ran `pnpm package`
- started `ferdium-app` locally with `pnpm debug`
- waited until I was told that services have been updated by prompt
- reloaded `google-calendar` service and dark mode works again

image

***

I think though that dark mode is always on now with how its set up. We would need to do something like in https://github.com/ferdium/ferdium-recipes/blob/main/recipes/android-messages/webview.js#L43-L47 I assume

mcmxcdev:
I updated the code to be able to toggle between rather than always have dark mode enforced in  [`23bf649` (#402)](https://github.com/ferdium/ferdium-recipes/pull/402/commits/23bf6499abf5b9d002deff1cf7aa4dfd6d08450f)

Here is a video showcasing it working:

[Screencast from 2023-08-07 19-28-35.webm](https://github.com/ferdium/ferdium-recipes/assets/16797721/90ce7afe-2b9f-47e2-89be-27e014bd7eb0)

lukassr:
It is important to note that, if enabling Dark Mode in an existing Google Calendar service does not work, you may need to delete it and create a new one.

PR #407, opened 2023-08-23
Fix telegram deeplink by adding empty onload

SpecialAro:
#### Pre-flight Checklist

#### Description of Change
This is the weirdest fix I've ever done, but it seems like if we "override" somehow the `window.onload()` we fix the issue described on https://github.com/ferdium/ferdium-app/issues/1275 (came upon this while trying to solve the issue). Basically, the problem is that whenever we click a button we push a new route that has a script to check for the window.protoUrl and then if it starts with `t.me` tries to navigate to the `tg://` deeplink. Seems like by introducing the `window.onload` method we somehow fix the behaviour, preventing it from navigating to deeplink.

Would be good if someone also tested this PR before merging as this fix is kind of weird.

### DIFF
--- a/recipes/telegram/package.json
+++ b/recipes/telegram/package.json
@@ -1,7 +1,7 @@
 {
   "id": "telegram",
   "name": "Telegram",
-  "version": "3.4.1",
+  "version": "3.4.2",
   "license": "MIT",
   "config": {
     "serviceURL": "https://web.telegram.org",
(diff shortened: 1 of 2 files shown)


vraravam (approved):
I haven't tested this locally

vraravam:
Confirmed that this is working in the latest build of Ferdium along with this fix.

PR #408, opened 2023-08-25
Fix Whatsapp recipe unread counts for muted conversations

gmurdocca:
…ions.

#### Pre-flight Checklist

#### Description of Change

This is a bugfix for [issue #1303](https://github.com/ferdium/ferdium-app/issues/1303). Unread counts/badges were erroneously incrementing for Whatsapp conversations that were marked as muted. When mute option "Always" is chosen in Whatsapp Web, the value of `chat.muteExpiration` in webview.js is set to `-1`, however the conditional that governs incrementing of unread counts expected a positive non-zero value in this variable to indicate a muted conversation. Chaging the comparison from `> 0` to `!= 0` solved the issue.

### DIFF
--- a/recipes/whatsapp/package.json
+++ b/recipes/whatsapp/package.json
@@ -1,7 +1,7 @@
 {
   "id": "whatsapp",
   "name": "WhatsApp",
-  "version": "3.5.1",
+  "version": "3.5.2",
   "license": "MIT",
   "config": {
     "serviceURL": "https://web.whatsapp.com",
--- a/recipes/whatsapp/webview.js
+++ b/recipes/whatsapp/webview.js
@@ -18,7 +18,7 @@ module.exports = Ferdium => {
       query.onsuccess = event => {
         for (const chat of event.target.result) {
           if (chat.unreadCount > 0) {
-            if (chat.muteExpiration > 0 || chat.isAutoMuted) {
+            if (chat.muteExpiration != 0 || chat.isAutoMuted) {
               unreadMutedCount += chat.unreadCount;
             } else {
               unreadCount += chat.unreadCount;


victorbnl:
Did you test it? I didn’t but I’m not sure it would work.

If a chat is unmuted, then `chat.muteExpiration` will be set to `undefined`. With the previous code, it did the comparison `undefined > 0`, which was `false`. So unmuted chats were indeed considered unmuted. Now, it will do `undefined != 0`, which will be `true`. This will make unmuted chats be considered muted.

Note: I’m not sure all unmuted chats’ `muteExpiration` attributes are set to `undefined`, but as far as I remember at least some of them are.

gmurdocca:
Thanks for the insight @victorbnl. The patch is tested working on Version: 6.5.0-nightly.2 (rpm release deployed to Fedora Linux 38). Tests were simple however, observing the unread count change as expected when conversations were muted and unmuted, and correctly aggregating unread counts in other unmuted conversations.

If Whatsapp Web behavior is intermittent (where only some chats yield a value of undefined), then perhaps the following update to the conditional would be an improvement:

```
if (chat.muteExpiration == -1 || chat.muteExpiration > 0 || chat.isAutoMuted) {
```

victorbnl:
Well, I just checked again on WhatsApp and it seems that all my chats do have a `muteExpiration` and a `isAutoMuted` set to `0` and `false`, respectively. Thus maybe it would require some further testing to make sure these attributes are never undefined. I just remember having issues with it (perhaps with an old version), but maybe I don’t remember well, I should look for the issue.

Though if we were to admit I am correct, that fix looks good to me.

gmurdocca:
Let me know if you prefer the updated conditional which covers the case that an errant undefined value appears for `muteExpiration` and I will update the PR.

victorbnl:
For me you can let it as it is. Testing seems to show the attributes are present. Let’s wait for feedback from the developers.

SpecialAro:
Hi guys! This is looking good :D Just tested and it works fine. I'm approving and merging this PR. Thank you @gmurdocca for your contribution

SpecialAro (approved):

SpecialAro:
@all-contributors - please add @gmurdocca for code

PR #420, opened 2023-09-06
Recipes maintenance

Arthur-Huan:
#### Pre-flight Checklist

#### Description of Change

Update icons for Snapchat, Evernote, Bitwarden, TickTick, Fastmail, Google Voice, Tinder, and Zenhub. Update the name and icon of Hangouts to Google Chat.


mcmxcdev (commented):
Wow, thanks a lot for all these great updates!

Could you try to scale down the Bitwarden and Tinder logos? They are big in size currently for no good reason, I guess 150px like the others would be ideal.

Also, I wonder if we can simply rename hangouts to google-chat since we had issues in the past with recipes not receiving any updates if their name changes. Wdyt @vraravam ?

Arthur-Huan:
@mcmxcdev Thanks, I didn't notice that they were so big. I also hadn't thought about the ramifications of simply renaming the recipes. That's a very good point.

I do think something should be done about it, especially since Google Hangouts is advertised very prominently in both the app description and as a “Ferdium pick” in the recipes page. Just don't know the proper way to do it rn.

vraravam:
Changing the name of a recipe will mean that the pre-existing installations will not get a new updated version. The user will then have to delete that service, and re-add/re-configure it on their local installation of Ferdium

vraravam (approved):

PR #425, opened 2023-09-23
feat: support for the new slack ui

mcmxcdev:
#### Pre-flight Checklist

#### Description of Change
- update selector for team menu to work with new and old slack ui
- bump slack recipe to v1.6.0

References https://github.com/ferdium/ferdium-app/issues/1371

#### Screenshots

Slack service has the icon again:
Screenshot from 2023-09-23 09-29-10-obfuscated

### DIFF
--- a/recipes/slack/webview.js
+++ b/recipes/slack/webview.js
@@ -38,12 +38,23 @@ module.exports = Ferdium => {
   const getTeamIcon = function getTeamIcon(count = 0) {
     let countTeamIconCheck = count;
     let bgUrl = null;
-    const teamMenu = document.querySelector(
+    // INFO: A new Slack UI was introduced in August 2023 and will be rolled out gradually,
+    // therefore we need to support both old and new UI for the time being
+    // See more: https://slack.com/blog/productivity/a-redesigned-slack-built-for-focus
+    const oldSlackUiTeamMenu = document.querySelector(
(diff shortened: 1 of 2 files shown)


recipes/slack/webview.js
```diff
@@ -38,12 +38,23 @@ module.exports = Ferdium => {
...
+    const newSlackUiTeamMenu = document.querySelector(
+      '.p-ia4_home_header_menu__button',
+    );
+
+    if (oldSlackUiTeamMenu || newSlackUiTeamMenu) {
+      if (oldSlackUiTeamMenu) {
+        oldSlackUiTeamMenu.click();
+      } else if (newSlackUiTeamMenu) {
+        newSlackUiTeamMenu.click();
+      }
```
SpecialAro, line 56: Haven't tested it (I don't use slack). But instead of this, can't we use an array of strings on which we pass all the query selectors (old and new)? Given that the logic for both new and old just vary on the query selector?

mcmxcdev: Not sure how that would look like, feel free to push a commit simplifying this!

I wasn't planning on keeping this code around for the long run and remove the old slack UI code in a couple of months.

SpecialAro: Lets go with your fix and then change it in a couple of months (when the new UI sticks) 😄 Thank you!

PR #437, opened 2023-10-10
Make the validation checks more noticeable

Arthur-Huan:
#### Pre-flight Checklist

#### Description of Change

Put the validation checks in a more noticeable place.

I've noticed a lot of PRs that fail the checks. A lot of the time, the error message says it's a `pnpm package` error message asking for a version bump, even when the `package.json` already has a version bump. It's usually because the two other commands edited the files when they were run, and if the changes weren't staged ahead of time locally, `pnpm package` sees the new changes and fails the build.

It seems that a lot of people either didn't read the validation checks in the docs, or forgot (as I did with my latest PR). So, I'm thinking it might be a good idea to tell/remind people of the validation checks and put the info in a more prominent place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,9 @@
 
 Recipes are small scripts that are responsible for providing the connection between your services (e.g. WhatsApp, Gmail or Slack) and Ferdium. It provides Ferdium information like the number of current notifications, handles enabling dark mode and may otherwise improve your experience with the service.
 
-## Creating and adding your own recipes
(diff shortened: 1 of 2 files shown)


Alphrag:
> Should we make a script in package.json that runs all 3 commands? so the user only has to run `pnpm validate` ?

Yeah completely agree with that! It would make it easier, and then in the PR template, we can add a checkbox "Have you run `pnpm validate`?" to remind the user.

PR #439, opened 2023-10-15
Fix message notifications in Telegram

nikolaykhodov:
#### Pre-flight Checklist

#### Test evidence

Here is a [recorded screencast](https://youtu.be/Jc-QQZw5QVE) how I sent a message to myself and a notification appears in the top-right corner.

#### Description of Change

I've identified the reason why Telegram Web A's notifications don't make it through in Ferdium:

Telegram A [sends notifications to the service worker](https://github.com/Ajaxy/telegram-tt/blob/8fc3df855df4a1d9914e90e9e4ac8c85c2d3dd61/src/util/notifications.ts#L484-L500) to handle them if push notifications are supported:

```js
  if (checkIfPushSupported()) {
    if (navigator.serviceWorker?.controller) {
      // notify service worker about new message notification
      navigator.serviceWorker.controller.postMessage({
        type: 'showMessageNotification',
        payload: {
          title,
          body,
          icon,
          chatId: chat.id,
          messageId: message.id,
          shouldReplaceHistory: true,
          isSilent: message.isSilent,
          reaction: activeReaction?.reaction,
        },
      });
    }
```

This condition is coded in the source code with the [following logic](https://github.com/Ajaxy/telegram-tt/blob/8fc3df855df4a1d9914e90e9e4ac8c85c2d3dd61/src/util/notifications.ts#L51):

```js
function checkIfPushSupported() {
  if (!IS_SERVICE_WORKER_SUPPORTED || IS_ELECTRON) return false;
```

However, the service worker encounters an issue when retrieving a list of notifications, resulting in [showNotification()](https://github.com/Ajaxy/telegram-tt/blob/8fc3df855df4a1d9914e90e9e4ac8c85c2d3dd61/src/serviceWorker/pushNotification.ts#L270) never being called:

```js
  if (e.data.type === 'showMessageNotification') {
    // store messageId for already shown notification
    const notification: NotificationData = e.data.payload;
    e.waitUntil((async () => {
      // Close existing notification if it is already shown
      if (notification.chatId) {
        const notifications = await self.registration.getNotifications({ tag: notification.chatId });
        notifications.forEach((n) => n.close());
      }
      // Mark this notification as shown if it was handled locally
      shownNotifications.add(notification.messageId);
      return showNotification(notification); // <-- THIS LINE
```
The root cause appears to be [self.registration.getNotifications()](https://github.com/Ajaxy/telegram-tt/blob/8fc3df855df4a1d9914e90e9e4ac8c85c2d3dd61/src/serviceWorker/pushNotification.ts#L265C35-L265C102) that returns a non-resolving promise:

```js
const notifications = await self.registration.getNotifications({ tag: notification.chatId });
```

This might be due to an Electron bug, although I didn't dive deeper into its specifics.

A solution might be to fix detecting whether [the runtime is Electron](https://github.com/Ajaxy/telegram-tt/blob/8fc3df855df4a1d9914e90e9e4ac8c85c2d3dd61/src/util/windowEnvironment.ts#L39) by instrumenting `window` with the `electron` property. This modification can only be performed from `index.js`, which has the ability to inject a JavaScript snippet prior to the loading of Telegram Web A. Initiating this from `webview.js` is not feasible as, typically, an app in the webview has already been initialized.

### DIFF
--- a/recipes/telegram/index.js
+++ b/recipes/telegram/index.js
@@ -9,12 +9,12 @@ function instrumenEnvironment(webview) {
     })();
   `);
 }
-module.exports = (Ferdium) =>
+module.exports = Ferdium =>
   class Telegram extends Ferdium {
     // https://www.electronjs.org/docs/latest/api/webview-tag/#dom-events
     get events() {
       return {
-        "load-commit": "loadCommit",
+        'load-commit': 'loadCommit',
       };
     }
 
--- a/recipes/telegram/index.js
+++ b/recipes/telegram/index.js
@@ -9,8 +9,8 @@ function instrumenEnvironment(webview) {
     })();
   `);
 }
-module.exports = (Franz) =>
-  class Telegram extends Franz {
+module.exports = (Ferdium) =>
+  class Telegram extends Ferdium {
     // https://www.electronjs.org/docs/latest/api/webview-tag/#dom-events
     events = {
       "load-commit": "loadCommit",
--- a/recipes/telegram/index.js
+++ b/recipes/telegram/index.js
@@ -12,9 +12,12 @@ function instrumenEnvironment(webview) {
 module.exports = (Ferdium) =>
   class Telegram extends Ferdium {
     // https://www.electronjs.org/docs/latest/api/webview-tag/#dom-events
-    events = {
-      "load-commit": "loadCommit",
-    };
+    get events() {
+      return {
+        "load-commit": "loadCommit",
+      };
+    }
+
     loadCommit(event) {
       instrumenEnvironment(event.target);
     }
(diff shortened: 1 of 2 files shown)


recipes/telegram/index.js
```diff
@@ -1 +1,21 @@
...
+        if(window.electron) {
+            return;
+         }
+        
+        window.electron = { };
+    })();
+  `);
+}
+module.exports = (Franz) =>
+  class Telegram extends Franz {
```
SpecialAro, line 13: Can we use `Ferdium` and extend `Ferdium` please?

SpecialAro (changes requested):

recipes/telegram/index.js
nikolaykhodov: Updated in https://github.com/ferdium/ferdium-recipes/pull/439/commits/172252a27ba8b60aaabb4d126b82ad719bfcd8f8

SpecialAro:
@nikolaykhodov I'm rebasing the branch to the latest code so you can run `pnpm validate` and commit the changes it does (this should fix the builds).

SpecialAro (approved):
LGTM! Haven't tested given that I can't message myself on telegram... But your video is enough!

Thank you for your contribution!

SpecialAro:
@all-contributors please add @nikolaykhodov for code

PR #445, opened 2023-10-21
feat: add memos recipe

lincolnthalles:
#### Pre-flight Checklist

#### Description of Change

Add recipe for [memos](https://github.com/usememos/memos)


SpecialAro:
Hi @lincolnthalles ! 

Thank you for logging this PR 😁

Just to clarify:
* Memos is self-hosted only, right? Meaning there is no "official" app rather than the demo?

If that is the case then let me suggest trying to change the `package.json` to something like the `Slack` recipe (that includes teamId and the teamIdSuffix). Can we test if making teamIdSuffix empty and the URL the full teamId works?

This way users will be forced to indicate a URL. Otherwise, if that doesn't work, we can keep it like is 😁

Let me know if you prefer that I test my suggestion on my side first!

lincolnthalles:
> Hi @lincolnthalles !
> 
> Thank you for logging this PR 😁
> 
> Just to clarify:
> 
> * Memos is self-hosted only, right? Meaning there is no "official" app rather than the demo?
> 
> If that is the case then let me suggest trying to change the `package.json` to something like the `Slack` recipe (that includes teamId and the teamIdSuffix). Can we test if making teamIdSuffix empty and the URL the full teamId works?
> 
> This way users will be forced to indicate a URL. Otherwise, if that doesn't work, we can keep it like is 😁
> 
> Let me know if you prefer that I test my suggestion on my side first!

It's self-hosted only, the domain can be anything. So, there's no point in using the subdomain template from Slack.
By just removing `serviceURL` Ferdium will require user to input a URL.

mcmxcdev:
Thanks for the added recipe @lincolnthalles !

Alphrag:
@all-contributors please add @lincolnthalles for code

PR #463, opened 2023-11-15
Change name of Tutanota to Tuta Mail

noraj:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

#### Description of Change

tuta renamed and changed domain

image

image

PR #423, opened 2023-09-15
Renamed Twitter recipe to X. Added a new icon & updated package.json

hafiz-muhammad:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [x] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md) for this project.
- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CODE_OF_CONDUCT.md) that this project adheres to.

PR #464, opened 2023-11-18
feat: update Tutunota, rebranding as Tuta

rdwz:
- [ ] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-recipes/blob/HEAD/CONTRIBUTING.md)* for this project.

*404 - page not found

#### Description of Change

Tutanota got rebranded as Tuta

This PR includes:
- name changed
- alias added
- icon replaced
- serviceUrl updated
- id unchanged


rdwz:
Todo
- [ ] open migration in ferdium-app instead of browser

according to [#1445](https://github.com/ferdium/ferdium-app/issues/1445)

There is an option to migrate saved credentials and a link is shown to do so, but
- it wont't work when opened in browser
- needs to be opened in ferdium-app

recipes/tutanota/package.json
```diff
@@ -1,9 +1,13 @@
 {
   "id": "tutanota",
-  "name": "Tutanota",
-  "version": "1.3.0",
+  "name": "Tuta",
+  "version": "1.4.0",
+  "description": "Tuta Mail",
+  "aliases": [
+    "Tutanota"
```
vraravam, line 7: If the recipe name is still `tutanota` (and this is valid/necessary to be able to upgrade automatically for those users who have already configured this in their installation of Ferdium), then the alias can be `tuta`. Having the alias as the same name as the recipe is redundant.

vraravam: If your intention is for the partial string match to kick in and match on "tuta," then the alias is unnecessary

vraravam (changes requested):

recipes/tutanota/package.json
mydarkstar: I think the alias is useful in case someone types out "Tutanota" in full or copy-pastes the name.
Or even just as a reference to the old name shown on the "Available services" screen.

We did the same thing here:
https://github.com/ferdium/ferdium-recipes/pull/423/files

rdwz: > If your intention is for the partial string match to kick in and match on "tuta," then the alias is unnecessary

I tried that, but it didn't work.

Maybe someone can help. How can I debug this script. However in the end this isn't nessessary to migrate credentials, just manual login and done.

mydarkstar:
If you don't mind, I would prefer merging #463 instead, as it is pretty much the same and does not change the logo, as described in the PR comments (Tuta just inverts all colors for dark mode).

mydarkstar:
Thanks for the help, Marc! I appreciate it lots.
I merged the previous PR of this, and you should see the service name update shortly!

We're also updating the broken link you noticed in the PR template in #469.
Have a good evening!

Duplicate of #463

rdwz:
> If you don't mind, I would prefer merging #463 instead, as it is pretty much the same and does not change the logo, as described in the PR comments (Tuta just inverts all colors for dark mode).

But this is the look of Tuta Mail, there is no other light mode with the old colors. All screenshots advertising the service show this new neon green look: https://tuta.com

mydarkstar:
> there is no other light mode with the old colors

There are three color schemes available: Light (red on white), Dark (inverted colors), and Blue (blue on white).
The Dark mode simply inverts the colors and this is not the "new" Tuta logo. It has also inverted the colors long before the rebranding, therefore this is not a new change.

> All screenshots advertising the service show this new neon green look

If you scroll down on the main website you linked and select the "Contacts" screenshot, it is shown in Light mode, with the non-inverted colors:
Tuta.com Contacts screenshot

I appreciate your input on this matter! However, I have to disagree, as there was definitely no logo change.
The way Tuta chose to implement the Dark mode in their products is unfortunate.

If you would like to use the inverted logo out of preference, do set a custom icon for your service of course!

PR #482, opened 2023-12-19
Use cache for lint scripts

rdwz:
#### Description of Change

> $ eslint --help
> ...
> Caching:
>   --cache                          Only check changed files - default: false
> ...

> $ prettier --help
> ...
>  --cache                  Only format changed files. Cannot use with --stdin-filepath.
>                           Defaults to false.


.gitignore
```diff
@@ -1,12 +1,13 @@
...
-node_modules
-archives/
-yarn-error.log
+.eslintcache
+.idea
 .pnpm-debug.log
-server*.log
-*.png
-*.jpg
-*.jpeg
```
SpecialAro, line 10: any reason for some of this deletion ?

SpecialAro (changes requested):

vraravam (commented):
I dont agree with the caching. We (as maintainers) have seen innumerable times where a merged code was also not linted/formatted properly. So, only processing the changed files will not take that into account. Also, the time taken for processing all files (as opposed to only the changed files) is quite bearable for now.

.gitignore
vraravam: i think the deleted lines are just moved.

vraravam:
Closing

PR #517, opened 2024-03-03
Hides the sidebar of distractions

Tomurisk:
That is only my suggestion, so I don't pressure to have this accepted, however "Instagram Direct Messages" service could work in a more similar manner to Distraction Free Instagram app, whereas those that want full features could add the regular Instagram service instead.

### DIFF
--- a/recipes/instagram-direct-messages/service.css
+++ b/recipes/instagram-direct-messages/service.css
@@ -11,3 +11,7 @@
 #react-root>section .i0EQd {
     max-width: unset !important;
 }
+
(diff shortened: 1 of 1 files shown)


mcmxcdev:
Can you provide a screenshot of what exactly you are hiding here?

What you consider as distraction might be important for someone else, so you could hide this on your end only too.

vraravam:
Closing since the OP hasn't answered the outstanding question and we are not sure of what's trying to be accomplished.

PR #581, opened 2024-08-22
Injecting `darkmode.css` for Messenger when appropriate

Arthur-Huan:
#### Pre-flight Checklist

#### Description of Change

This is mostly a proof of concept.

Currently, none of the services display their dark themes even with the system in dark mode, Ferdium in dark mode, and the universal dark mode feature turned on. It's long bothered me that some recipes have `darkmode.css` files but don't seem to be injected anywhere. So I tested out making some changes to `webview.js`, to inject the CSS if it's detected that the system is in dark mode. It seems to work. I am proposing that the previous snippet be added to all recipes if they have a `darkmode.css` file.

To me, it seems this would solve the dark mode issue (assuming we have a functional `darkmode.css` file). But I do wish to be cautious and see what more senior and knowledgeable contributors think before I proceed.

### DIFF
--- a/recipes/messenger/webview.js
+++ b/recipes/messenger/webview.js
@@ -127,8 +127,10 @@ module.exports = (Ferdium, settings) => {
     true,
   );
 
-  if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
+  if (
+    window.matchMedia &&
+    window.matchMedia('(prefers-color-scheme: dark)').matches
+  ) {
     Ferdium.injectCSS(_path.default.join(__dirname, 'darkmode.css'));
   }
-
 };
--- a/recipes/messenger/webview.js
+++ b/recipes/messenger/webview.js
@@ -126,4 +126,9 @@ module.exports = (Ferdium, settings) => {
     },
     true,
   );
+
+  if (window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches) {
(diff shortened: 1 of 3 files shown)


Arthur-Huan:
Added note:

The original `darkmode.css` for Messenger doesn't work at all. I stole one from Instagram. It also doesn't work that well, but at least you can see some changes to show that the CSS is being injected properly during dark mode, and not injected during light mode. As a proof of concept.

It seems like most of the `darkmode.css` files would need changes as I've struggled finding one that works perfectly. Perhaps they're outdated as they seem to have just been copied from Franz and not updated since. That would be what I look at next, to see if I can update and fix them.

recipes/messenger/webview.js
```diff
@@ -126,4 +126,11 @@ module.exports = (Ferdium, settings) => {
     },
     true,
   );
+
+  if (
+    window.matchMedia &&
+    window.matchMedia('(prefers-color-scheme: dark)').matches
```
SpecialAro, line 132: We should not be using this but instead use the internal config of the service for handling dark mode (I'm on mobile right now but should be something like service.isDarkMode)

SpecialAro (commented):
Hello 👋 

Thank you for opening this PR!

Nevertheless, we should not be using the prefers dark mode but the internal config for the service in order to inject the dark mode.

vraravam (changes requested):
same comment as @SpecialAro

PR #600, opened 2024-10-16
Adding support for Micosoft Windows

Justman100:
#### Pre-flight Checklist

- [ ] I updated the version package [Updating](https://github.com/ferdium/ferdium-recipes/blob/main/docs/updating.md#4-updating-the-version-number) 

#### Description of Change
Adding support for Microsoft Windows in the recipe-create-script

PS: I leave the version upgrade to the maintainers

### DIFF
--- a/scripts/create.mjs
+++ b/scripts/create.mjs
@@ -43,11 +43,14 @@ const pascalCasedName = toPascalCase(recipe); // PascalCased recipe ID only cont
 (async () => {
   // Folder paths
   const userData =
-  process.env.APPDATA || (
+  process.env.APPDATA || (  
+    // Check if this script runs on Windows
     process.platform === 'win32'
       ? `${process.env.USERPROFILE}\\AppData\\Roaming\\Ferdium\\recipes\\dev`
+    // If not, check for Darwin
     : process.platform === 'darwin'
       ? `${process.env.HOME}/Library/Application Support`
+    // If fails both checks, simply use this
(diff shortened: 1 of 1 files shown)


scripts/create.mjs
```diff
@@ -43,10 +43,13 @@ const pascalCasedName = toPascalCase(recipe); // PascalCased recipe ID only cont
 (async () => {
   // Folder paths
   const userData =
-    process.env.APPDATA ||
-    (process.platform === 'darwin'
+  process.env.APPDATA || (
```
vraravam, line 46: Please do not use nested ternary syntax. Please change to either `switch-case` statement or `if-else` conditions.

vraravam (changes requested):

scripts/create.mjs
Justman100: But... It works!

vraravam: the question is not only about it working, its also about understabability/maintainability

Justman100: What do you mean? Those who are not familiar with it will not look at the source code or try to set it up! One should have a certain amount of experience if one want to work in this area! And if one have that...

Justman100: But please, have added notes! I think that should be enough.

vraravam: maintainability

vraravam (approved):

PR #613, opened 2025-01-26
franz-custom-website: fix click events when trapLinkClicks is enabled

s-ol:
#### Pre-flight Checklist

#### Description of Change

Fully deactivates messing with `click` events when "Open external URLs in Ferdium" is enabled. Previously Ferdium would eat all click events on `<a href="http://...` and `<button title="http://...` tags, breaking sites that override click events to provide in-page navigation.

Close ferdium/ferdium-app#2028

### DIFF
--- a/recipes/franz-custom-website/package.json
+++ b/recipes/franz-custom-website/package.json
@@ -1,7 +1,7 @@
 {
   "id": "franz-custom-website",
   "name": "Custom Website",
-  "version": "1.5.1",
+  "version": "1.6.0",
   "license": "MIT",
   "repository": "https://github.com/meetfranz/recipe-custom-website",
   "config": {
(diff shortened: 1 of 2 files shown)


Alphrag:
Hi @s-ol, the use case you are mentioning should not warrant for a complete suppression of this feature, however we would be happy to see the code improved. The problem described in the issue linked is about the question of whether the event should be stopped from propagating or not. Indeed, on some websites if this event is not stopped when trapLink is on, then the webpage will open in the browser anyway, which goes against what the goal of trapLink was in the first place. 
So if there is a way to control than when this were to happen we block the event from happening, and we let it be if it is instead helping to move in the page, then that would be a perfect change. At the current status, we do not want to merge such a change that simply disable it entirely.

PR #626, opened 2025-03-17
Update MS Teams serviceURL

infinitnet:
Old serviceURL redirects to new serviceURL so should probably be updated.

#### Pre-flight Checklist

#### Description of Change

teams.microsoft.com (old serviceURL) redirects to teams.live.com (new serviceURL).

### DIFF
--- a/recipes/msteams/package.json
+++ b/recipes/msteams/package.json
@@ -1,13 +1,13 @@
 {
   "id": "msteams",
   "name": "Microsoft Teams",
-  "version": "3.3.2",
+  "version": "3.3.3",
   "license": "MIT",
   "aliases": [
     "teamsChat"
   ],
   "config": {
-    "serviceURL": "https://teams.microsoft.com",
+    "serviceURL": "https://teams.live.com",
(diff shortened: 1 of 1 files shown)


SpecialAro:
Hello! Thank you for raising this PR.

In fact, I don't think that `teams.microsoft.com` is an old URL. This URL is specific for O365 users whether `teams.live.com` is for common users that are not under the 365 package offer.

This issue is transversal to all microsoft services. Someone tried to do the same change as you propose here for another service (https://github.com/ferdium/ferdium-recipes/pull/353) but we didn't merge it because another solution is needed to solve this issue - given the fact that we were only be changing the affected users of this bug.

I propose that, if you got the time and knowledge, to look at the `ferdium-app` codebase and try to code a solution that is transversal to all microsoft services.

infinitnet:
@SpecialAro Oh, thanks for the clarification - I had no idea! :)

Considering the current implementation is actually working fine and I'm not good enough of a dev to figure out a transversal solution, I'll close the PR.

PR #628, opened 2025-03-18
Add Miniflux recipe

makinori:
#### Pre-flight Checklist

#### Description of Change

Added a recipe for Miniflux RSS reader

https://miniflux.app/


makinori:
I don't know if I should add this to webview.js, but it might  be a good idea to reload the page occasionally as Miniflux doesn't automatically fetch new posts. Any ideas?

Alphrag:
That is a very good point about the auto-refresh. However, I would not go with it that way, since it means that if a user is browsing something, the reader to refresh automatically on them and they have to search again what they were reading.
I guess that users should then either refresh manually when they are on it and want to check if there are any new elements, or use the service with the hibernation strategy, since then it will indeed reload from time to time without hindering what they are doing. What do you think?

makinori:
Is there a way to check if the user is currently looking at the service? Could use that to prevent the refresh.

Alphrag:
That could be an idea indeed. At the moment, there is no way from a service to know if it is in view. The way I see it (but I don't know the whole codebase yet), you would need to modify the app itself, in order to add a function in `RecipeWebview.ts`, say `isServiceOnFocus(serviceId)`, that you would then call in the webview of this recipe to do what you need about this.
Now the actual implementation of might need to use the `ipcRenderer` to send an event to the main app, from where you will then be able to check the `isActive` value of the Service this belongs to (potentially reading it from `this.ferdium.stores.services.active.id`).

makinori:
Was able to figure out a way to do this using `blur` and `focus` events. If you open the service dev tools and reload whilst focused or unfocused, the timeout will set accordingly and adjust when active changes.

makinori:
I realized it would make more sense to only reload when the user isn't reading an entry or in settings

SpecialAro:
@makinori I apologize that I just realized now we merged https://github.com/ferdium/ferdium-recipes/pull/655 before this PR.
Nevertheless, I'm giving you credit for working on this and I kindly ask that, if you think it is still valid, you raise a new PR to add the auto-refresh feature.

I'm closing this as won't merge but I would really appreciate if you could open a new PR with that change 🙏

SpecialAro:
@all-contributors please add @makinori for code

PR #634, opened 2025-03-28
Add recipe for Microsoft Teams Personal (teams.live.com)

krisrok:
#### Pre-flight Checklist

- [ ] I updated the version package [Updating](https://github.com/ferdium/ferdium-recipes/blob/main/docs/updating.md#4-updating-the-version-number) => _No, this is mostly a copy of an existing recipe_

#### Description of Change

Having both 365-teams (teams.microsoft.com) and a "Personal" team (teams.live.com) under the same account works for the 365-teams. When reloading the personal team though we have to confirm logging into the personal team again every time.

I added a copy of the msteams recipe and changed the base url to teams.live.com.

This way, the service now correctly logs in into the personal team without needing further interactions.

### DIFF
--- a/recipes/msteams-personal/index.js
+++ b/recipes/msteams-personal/index.js
@@ -1,0 +1,21 @@
+module.exports = Ferdium =>
+  class MicrosoftTeams extends Ferdium {
+    overrideUserAgent() {
+      return window.navigator.userAgent
+        .replaceAll(/(Ferdium|Electron)\/\S+ \([^)]+\)/g, '')
+        .trim();
+    }
+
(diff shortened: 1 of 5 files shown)


Alphrag:
Hi @krisrok, I am not a fan of duplicating the whole recipe only to change the url, because this will cause problems maintaining it. 
I would suggest instead modify the current `Teams` recipe so that it has a possible default to `teams.microsoft.com`, but allow the user to enter a more custom url such as `teams.live.com`. A good example for that is the Office 365 outlook recipe, which works in the same way.

PR #649, opened 2025-09-13
Add Copilot recipe

kurobeats:
#### Pre-flight Checklist

- [ X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [ X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [ X] I updated the version package [Updating](https://github.com/ferdium/ferdium-recipes/blob/main/docs/updating.md#4-updating-the-version-number) 

#### Description of Change

Alrighty! A Copilot recipe to address the request in Issue https://github.com/ferdium/ferdium-app/issues/2002

It's super basic (I'm not javascript dev and thank the gods you don't demand it). I've tested it and done exactly what's required in the dev guide. I hope you accept it!


kurobeats:
Oh I made a new push, but the Syncthing one came through too. OK, well, just for all the users of Syncthing, a recipe with flexible URL support. I will add it's version 1 without notification support. I have no instances with error/notifications but as soon as I have a conflict I'll see if I can't add some notification count features.

Alphrag (changes requested):
Thanks for your contribution. However, there are quite a few changes that are required:
- revert the version change for eslint, removing the pnpm-lock file by the same occasion;
- except from the part that was commented out, keep `webview.js` intact from the creation (you can verify what it should look like with the one from the `sample_recipe` in the `script` folder)
- for Syncthing, it is better to use the `urlInputPrefix` with `teamID` as the placeholder for the port in `serviceURL` (since syncthing might run on different port depending on the service that launches it) rather than setting an url and allowing a custom one (which would prevent using it with the port 8080 because of how we understand setting both these values at the same time). If you really want to allow the possibility of entering a completely different url than `localhost`, then you could set the `hasHostedOption` option and only in this case setting `hasCustomURL` to true.
- for OPNsense, if I understand correctly, this is necessarily self-hosted, so you should remove the `serviceURL` and only set `hasCustomURL`. In the event where all the possible self-hosted urls look alike, you could otherwise do as above with setting both `urlInputPrefix` and `urlInputSuffix`, but this might not be the most intuitive version for user that wish to use this recipe.
- for both Syncthing and OPNsense, I would also suggest adding a field `message` to explain what needs to be filled in, just in case.

PR #665, opened 2026-01-09
Updated Microsoft Teams to the new URL

cedricroijakkers:
#### Pre-flight Checklist

Please ensure you've completed all of the following.

- [X] I have read the [Contributing Guidelines](https://github.com/ferdium/ferdium-app/blob/develop/CONTRIBUTING.md) for this project.
- [X] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/develop/CODE_OF_CONDUCT.md) that this project adheres to.
- [X] I updated the version package [Updating](https://github.com/ferdium/ferdium-recipes/blob/main/docs/updating.md#4-updating-the-version-number) 

#### Description of Change

This PR fixes issue: https://github.com/ferdium/ferdium-app/issues/2320

Microsoft is switching Teams to the new URL: https://teams.cloud.microsoft

People are slowly starting to get notifications that the is the "new" and "more secure" URL. So we need to switch over.

The new URL is already live and I am already using it in the browser.

PR #680, opened 2026-04-16
Handle new Microsoft Teams URL for isTeamsV2 detection #665

joschi127:
Improvement / fix for #665

#### Pre-flight Checklist

### DIFF
--- a/recipes/msteams/package.json
+++ b/recipes/msteams/package.json
@@ -1,7 +1,7 @@
 {
   "id": "msteams",
   "name": "Microsoft Teams",
-  "version": "3.3.4",
+  "version": "3.3.5",
   "license": "MIT",
   "aliases": [
     "teamsChat"
--- a/recipes/msteams/package.json
+++ b/recipes/msteams/package.json
@@ -1,7 +1,7 @@
 {
   "id": "msteams",
   "name": "Microsoft Teams",
-  "version": "3.3.3",
+  "version": "3.3.4",
   "license": "MIT",
   "aliases": [
     "teamsChat"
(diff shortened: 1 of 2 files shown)


Alphrag:
@joschi127 Well spotted. Please run the linter locally and push your changes, otherwise the build test fails because the file is modified in the process.

joschi127:
@Alphrag thanks for the quick reply. And sorry, I just did a quick edit directly on github.com without even checking out the code locally.  I did so now and pushed an update. Hope it's fine like this.

Alphrag:
The version should only be bumped by 1 tick for the whole PR. To make sure it does not complain locally, you actually need to make the change, commit and only then run the packaging check. If I remember correctly, it should then check relatively to the latest branching of the current branch, but it might sometimes require squashing some commits to make it pass locally, but it doesn't matter, as long as on GitHub the action runs fine (which compares the whole PR with the current main branch). 
Sorry that it's not fully clear on the contributing guidelines, we ought to improve them...

joschi127:
> The version should only be bumped by 1 tick for the whole PR. To make sure it does not complain locally, you actually need to make the change, commit and only then run the packaging check. If I remember correctly, it should then check relatively to the latest branching of the current branch, but it might sometimes require squashing some commits to make it pass locally, but it doesn't matter, as long as on GitHub the action runs fine (which compares the whole PR with the current main branch). Sorry that it's not fully clear on the contributing guidelines, we ought to improve them...

thx for explaining, update pushed

SpecialAro:
Thank you @joschi127  and @Alphrag! Merged it now.

@all-contributors please add @joschi127 for code

joschi127:
thx ❤️

PR #683, opened 2026-04-17
Add FluffyChat recipe

codeshell:
#### Pre-flight Checklist

## Adding new recipe

Service: FluffyChat Web

URL: https://fluffychat.im/web

Service ID: fluffychat

Recipe request issue in the main issue tracker : https://github.com/ferdium/ferdium-app/issues/2335

### Terminal output

```bash
pnpm i && pnpm lint:fix && pnpm reformat-files && pnpm package
...
recipes/fluffychat/index.js 1ms (unchanged)
recipes/fluffychat/package.json 1ms (unchanged)
recipes/fluffychat/webview.js 5ms (unchanged)
...
✅ Successfully packaged and added 410 recipes (0 unsuccessful recipes) in 71.992 seconds
```

### Additional information

- [x] I am the original creator of this package
- [x] I have run the `pnpm lint:fix && pnpm reformat-files && pnpm package` and verified that there are no validation errors reported for this package
- [x] My recipe has been tested to work inside Ferdium

### Details

This adds a new recipe for the webclient[^1] of FluffyChat[^2].

It can be used as an alternative to the deprecated[^3] SchildiChat recipe[^4].

Recipe works around flutter by reading from IndexedDB.

Related issues: https://github.com/ferdium/ferdium-app/issues/2335

[^1]: https://fluffychat.im/web
[^2]: https://github.com/krille-chan/fluffychat
[^3]: https://schildi.chat/legacy/
[^4]: [recipes/schildichat](https://github.com/ferdium/ferdium-recipes/tree/af8df5a255dd3692ffdb2f51be17c98190338529/recipes/schildichat)


andreclemente:
I've been testing it for a while now and so far it's been working great. The only issue I noticed is that OIDC doesn't work (because it's trying to opening a new window and Ferdium doesn't like it).

codeshell:
Thanks for the feedback, much appreciated!

I use it daily myself with two instances of this service configured as always active and I haven't experienced any memory leaks or cpu spikes.

In case you, have missed it. [FluffyChat 2.6.0](https://github.com/krille-chan/fluffychat/releases/tag/v2.6.0) was released and finally enabled **support for copy&paste of attachments/images**. I tested it and it also works within Ferdium 🎉 

As for social logins, you are pretty much putting the [Gretchen question](https://en.wikipedia.org/wiki/Gretchenfrage) on the table here for Ferdium. The main reason being that SSO / OIDC / OAuth and all the variants of official and custom implementations all conceptually work as the exact opposite of what Ferdium has been build for. That is:
- The former relaying authentication of one service to another completely different / "unrelated" service.
- The latter isolating different / "unrelated" services from each other.

### My thoughts on social logins from when I was creating the recipe and if/how I should fix it

This cannot (and should not) be resolved within recipes currently. I'm watching where Ferdium is heading with great interest myself. Until recently, I have only seen core hacks for some specific endpoints (like [Google SSO](https://github.com/ferdium/ferdium-app/pull/2261)) or ignored issues for others (like [Microsoft](https://github.com/ferdium/ferdium-app/issues/1128)).

A matrix client is the best example of why any endpoint-specific solution does not matter.

Depending on which [matrix] homeserver the user is trying to login, the login endpoint can literally be on any domain out there (e.g. self-hosted synapse with selfhosted authentik/keycloak/...).

For that reason there is no point in hacking the recipe for a specific service like Google as workaround.

If this was to be resolved in the recipe code, the service would have to allow a settings field where the user adds their custom authentication services that should be recognised / intercepted.

There might also have to be an indicator of how the implementation behaves. E.g. html link vs. javascript popup, shared cookie session vs url query parameter.

> For example, using FluffyChat with matrix.org SSO kind of has a manual workaround, because tokens are transmitted via url query params: matrix.org would open in your default browser, logging in would redirect to `https://fluffychat.im/web/auth.html?loginToken=<secret key>` in your default browser (doing nothing there). This could be manually copied back into your FluffyChat service (need to `View > Toggle Navigation Bar`) address bar and should continue the login process there.

Another aspect is FluffyChat being a WASM app, thus common DOM hacks from other recipes most likely wont work here.

### new development

I said "until recently", because I just found out there was a merge in May that looks like it could make a difference: [fix: open OAuth popups in-app to preserve window.opener for auth callbacks](https://github.com/ferdium/ferdium-app/pull/2412/) was merged after the latest release.

This might fix social logins for some login services. I'm eager to try it out once the next Ferdium release hits the road.


_This got a bit lengthier than expected, but it might be helpful to understand why I decided to not include any login related shenanigans into the recipe for now._

andreclemente:
Thanks for the detailed explanation. That actually makes a lot of sense.

The recently merged OAuth popup change sounds promising! It'll be interesting to see how much that improves things once the next release is out. 

In the meantime, I can confirm that the Matrix/FluffyChat SSO issue I was seeing is likely related to exactly the kind of flow you described.

Thanks again for maintaining the recipe and for sharing the background behind the design decisions. It's always helpful to understand not just what isn't implemented, but why.

SpecialAro (approved):

SpecialAro:
@all-contributors please add @codeshell for code

PR #700, opened 2026-06-30
Add CRM AI2N recipe

finesoft2009:
Add a new Ferdium recipe for CRM AI2N - a powerful AI-driven CRM system.

### DIFF
--- a/recipes/crm-ai2n/darkmode.css
+++ b/recipes/crm-ai2n/darkmode.css
@@ -1,0 +1,15 @@
+/* CRM AI2N Ferdium Dark Mode */
+body {
+  background-color: #1a1b1e !important;
+  color: #e0e0e0 !important;
+}
+
+input, textarea, select {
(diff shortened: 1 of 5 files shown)


SpecialAro (commented):
Thanks for the contribution! Overall, the recipe structure looks good, but I think a couple of points should be clarified before merging.

The main question is whether CRM AI2N actually supports custom/self-hosted URLs. If the recipe is only intended to use https://crm.ai2n.dev, then hasCustomUrl and validateUrl() are unnecessary and should probably be removed.

If custom instances are supported, then keeping hasCustomUrl makes sense, but validateUrl() should verify that the supplied URL is actually a CRM AI2N instance rather than only checking whether it responds successfully to a HEAD request.

Also, the unread badge parsing could use Ferdium.safeParseInt() to avoid a malformed badge value causing the total count to become NaN.

Also, please remove all the comments or at least translate them to english please

PR #702, opened 2026-07-16
feat(claude-ai): enable custom URL support

marek-knappe:
## Summary
- Enables `hasCustomUrl` for the Claude AI recipe so users can configure a custom landing URL
- Claude now offers multiple experiences at different paths (e.g. `claude.ai/code`, `claude.ai/design`) and users may want to land directly on one of them instead of the default `claude.ai` chat

## Test plan
- [x] Add Claude AI service with default URL — should still work as before (`https://claude.ai`)
- [x] Add Claude AI service with custom URL set to `https://claude.ai/code` — should land on Claude Code
- [x] Add Claude AI service with custom URL set to `https://claude.ai/design` — should land on Claude Design

### DIFF
--- a/recipes/claude-ai/package.json
+++ b/recipes/claude-ai/package.json
@@ -4,6 +4,7 @@
   "version": "1.0.0",
   "license": "MIT",
   "config": {
-    "serviceURL": "https://claude.ai"
+    "serviceURL": "https://claude.ai",
+    "hasCustomUrl": true
   }
 }
--- a/recipes/claude-ai/package.json
+++ b/recipes/claude-ai/package.json
@@ -1,7 +1,7 @@
 {
   "id": "claude-ai",
   "name": "claude.ai",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "license": "MIT",
   "config": {
     "serviceURL": "https://claude.ai",


SpecialAro (changes requested):
Thanks for the contribution! I like the idea of making Claude Code and Claude Design available in Ferdium, but I don't think `hasCustomUrl` is the best way to model this.

`hasCustomUrl` is intended for services where the URL is actually user-specific/custom, such as self-hosted or on-premise services. In this case, `https://claude.ai/code` and `https://claude.ai/design` are known, first-party Claude product endpoints rather than custom deployments.

I think it would be cleaner to add these as separate recipes instead:

* **Claude** → `https://claude.ai`
* **Claude Code** → `https://claude.ai/code`
* **Claude Design** → `https://claude.ai/design`

This would make them discoverable directly in Ferdium, avoid requiring users to know and enter the URLs themselves, and allow each service to have its own configuration/icon or service-specific handling in the future.

It also keeps the meaning of `hasCustomUrl` consistent with the existing recipe configuration API.

So I'd prefer separate recipes for Claude Code and Claude Design rather than enabling arbitrary custom URLs on the existing Claude recipe.

If you are willing to work on this, please open a separate PR for each recipe as I'm going to close this one.

Thank you 😄

marek-knappe:
The best would be to have select box as an option do u know if there is any plan to have that as config option?

PR #703, opened 2026-07-18
Add darkmode for Gmail incl. Gemini side panel

beto621:
## Description

Adds a `darkmode.css` for the Gmail recipe (it previously had none), providing a full dark theme when the service's dark mode is toggled.

Notable: Google's new **Gemini widget** (side panel + "Summarize this email" chip) forces the light GM3 token palette on its wrappers regardless of theme. Instead of chasing obfuscated per-element classes, this stylesheet re-declares Gmail's own official dark palette (`--gm3-sys-color-*`, copied from Gmail's shipped dark-theme rule) scoped to the Gemini wrappers, so every descendant — including future UI Google adds inside the panel — resolves dark automatically.

Also bumps the recipe version `1.6.3` → `1.6.4`.

## Testing

Verified live against mail.google.com: inbox, email reading view, compose, menus, search, the Gemini side panel, its suggestion chips, the Ask Gemini prompt box, and the summarize chip all render in proper dark colors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/recipes/gmail/darkmode.css
+++ b/recipes/gmail/darkmode.css
@@ -1,0 +1,660 @@
+* {
+  scrollbar-color: #383d3f #131617;
+}
+
+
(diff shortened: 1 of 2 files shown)


SpecialAro (changes requested):
Thanks for the contribution! The Gemini-specific changes look reasonable overall.

My main concern is the number of message-specific or dynamically generated selectors in `darkmode.css`, such as `.m_...`, `.msg...`, and IDs like `#\:rm`. These appear likely to come from particular emails or Gmail sessions rather than stable Gmail UI elements, which could make the stylesheet brittle and harder to maintain over time.

Could we remove or replace those generated/message-specific selectors with more stable Gmail selectors where possible?

SpecialAro:
image
Gemini side panel isnt working

PR #707, opened 2026-07-26
Gotify: badge always shows 1

### LINKED ISSUE #2459: Gotify Badge Always Shows 1
### Preflight Checklist

- [x] I agree to follow the [Code of Conduct](https://github.com/ferdium/ferdium-app/blob/HEAD/CODE_OF_CONDUCT.md) that Ferdium Recipes adheres to.
- [x] I have searched the [issue tracker](https://github.com/ferdium/ferdium-app/issues?q=is%3Aopen+is%3Aissue+label%3A%22recipe+%3Abookmark_tabs%3A%22+label%3A%22bug+%3Abeetle%3A%22+) for a bug report that matches the one I want to file, without success.

### Ferdium Version

7.1.2

### Recipe Name

gotify

### Recipe Version

1.1.1

_No response_

### Steps to reproduce

1. Add Gotify to Ferdium
2. Gotify needs at least 2 messages

### Expected Behavior

1. Ferdium displays a badge with the number of notifications loaded by Gotify

### Actual Behavior

Gotify in Ferdium always shows 1 on its badge.

### Screenshots

_No response_

### Additional Information

The existing Gotify recipe sets the badge count via a DOM based lookup of `#messages`.
`#messages` is always present - it's a wrapper/container for the individual messages, and will always return regardless of how many actual messages there are.

Gotify only renders DDM elements that are visible, so relying on the DOM elements will often report fewer messages than there actually are. For example, if you have 50 messages, and only 10 fit on your screen, counting the individual message elements will only report 10 instead of 50.

lreading:
#### Pre-flight Checklist

#### Description of Change

Closes https://github.com/ferdium/ferdium-app/issues/2459 

Gotify's existing recipe sets the badge count to reflect the count of `#messages`. That is a container only, and individual messages are children of that container. Because of this, the badge count is always 1.

Gotify uses React Fiber, and will only render DOM elements that are visible. Relying on visible DOM elements is inconsistent. If you have 50 messages, but only 10 are visible, the logic would show 10 instead of the real 50.

This implementation isn't perfect: this only reports the number of messages loaded by the Gotify client instead of using a direct API integration. Adding an API integration would require setting up an API key, or injecting into existing XHR requests, which feels a bit hacky. I think this loaded messages approach will be accurate more often than not.

This could become brittle if Gotify changes front-end frameworks, or even during major version upgrades of the front-end framework used by Gotify. I think this is a fair tradeoff.

This was manually tested in Ferdium 7.1.2 with 86 loaded messages, 0 messages, sending additional high-priority Gotify notifications. lint, reformat-files, and package all pass.

### DIFF
--- a/recipes/gotify/badge.js
+++ b/recipes/gotify/badge.js
@@ -1,0 +1,61 @@
+(() => {
+  if (window.gotifyFerdiumBadgeInjected) {
+    return;
+  }
+
+  window.gotifyFerdiumBadgeInjected = true;
+
+  const sendBadge = count => {
+    window.dispatchEvent(
+      new CustomEvent('gotify-ferdium-badge', {
+        detail: { count },
+      }),
+    );
+  };
+
+  const findReactFiber = element => {
+    const key = Object.keys(element).find(name =>
+      name.startsWith('__reactFiber$'),
+    );
+
+    return key ? element[key] : null;
+  };
+
+  const isMessageList = data =>
+    Array.isArray(data) &&
+    data.every(
+      message =>
+        message &&
+        typeof message.id === 'number' &&
+        typeof message.message === 'string' &&
+        typeof message.appid === 'number',
+    );
+
+  const getLoadedMessageCount = () => {
+    const root = document.querySelector('#messages');
+
+    if (!root) {
+      return 0;
+    }
+
+    // A Fiber's `return` property points to its parent Fiber. React sets it to
+    // null when the parent chain ends, so walk up the tree until that point.
+    let currentFiber = findReactFiber(root);
+
+    while (currentFiber !== null) {
+      const data = currentFiber.memoizedProps?.data;
+
+      if (isMessageList(data) && data.length > 0) {
+        return data.length;
+      }
+
+      currentFiber = currentFiber.return ?? null;
+    }
+
+    return 0;
+  };
+
+  window.addEventListener('gotify-ferdium-poll', () => {
+    sendBadge(getLoadedMessageCount());
+  });
+})();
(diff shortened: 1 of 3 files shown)


SpecialAro:
Hi! Thanks for your work on this.

I'm tempted to close this PR because #709 implemented it more easily. Would you be able to check and review that PR and see if everything is ok? I don't have a way of testing this since I don't use gotify...

lreading:
Hi! 

The TLDR is that #709 will only ever set the badge to a 1 or a 0. The badge does _not_ reflect the actual number of messages. Eg, if I have 200 messages, the badge is still set to 1.
This PR sets the badge to the number of messages (until you have multiple thousands, weird edge case due to Gotify internals).

So the decision between the two PRs, at least IMO, is based on the desired behavior. :smile: 

For what it's worth, the recipes I use set the badge to the number of alerts. I haven't seen it used as a binary indicator of you have alerts or you do not. There could be other recipes that do that though! Just not the ones I'm using today. :sunglasses: 

If you think the more binary badge count is the right move, then #709 is definitely the way to go!

A note on complexity - Gotify doesn't expose a reliable way of fetching the total message count, which is why this code is a bit weird. I'll admit that it's a bit of a hack - it's tapping into the React Fiber internals. :sweat_smile: 

If it helps, here's a couple screenshots showing the differences in behavior between the 2 approaches.

### Recipe from #709
_There are 14 total messages in Gotify - the badge only shows 1_.
image


### Recipe from this PR
_There are 14 total messages - the badge shows 14_.
image

The latter is what I would have expected it to function like. Buuuuut, I'm not the only person with opinions! Let me know how you want to proceed! :raised_hands:

SpecialAro (approved):
You are right! Thanks for taking the time to test both PRs 😄

SpecialAro:
@all-contributors please add @lreading for code

